PR #73, opened 2015-07-06
Add the user ID to the request log

chilts:


tokenserver/views.py
```diff
@@ -101,11 +101,14 @@ def valid_assertion(request):
     request.metrics['token.assertion.verify_success'] = 1
     request.validated['assertion'] = assertion
 
+    # pull out the FxA uid from the email address to log with the request
+    email = request.validated['assertion']['email']
+    request.metrics['id'] = email.split("@", 1)[0]
```
rfk: As much as I don't like services doing things their own special way, I wonder if it would be cleaner to treat tokenserver as a special case and just emit the entire "email" field here.

For historical reasons, the tokenserver code identifies users by their full "acct@provider" email-like address rather than just a uid.  IIRC this new code would be the only code in the whole program that assumes it's dealing with FxA rather than a generic email-based IdP.  If we set `request.metrics['email']` here rather than splitting off the uid, it will be more obvious what's going on and more inline with what the surrounding code does.  We can then fix it up in post-processing in heka.

Thoughts?

chilts: That sounds fine to me. Let's go with `email` and then heka can do it's thing there.

rfk:
No idea why we never merged this, but anyway, r+ :-)

rfk:
@chilts hiya! :-)

PR #75, opened 2015-10-20
Remove extra SQS message read

mostlygeek:
- nothing happens with this extra message read
- causes SQS to timeout on visibility and put it back on the queue
- causes hundreds of messages to pile up waiting for timeout on the
  queue


mostlygeek:
@rfk r? please

If it looks good, please merge and tag/release 1.2.16

rfk:
nice catch, r+

rfk:
>  tag/release 1.2.16

Tag created

PR #77, opened 2016-01-04
add fxa account reset event processing

dannycoates:
Handle `reset` events from the auth-server by updating the user's `generation`.

### DIFF
--- a/tokenserver/scripts/process_account_deletions.py
+++ b/tokenserver/scripts/process_account_deletions.py
@@ -80,7 +80,7 @@ def process_account_deletions(config_file, queue_name, aws_region=None,
                     if event_type == "delete":
                         # Mark the user as retired.
                         # Actual cleanup is done by a separate process.
-                        logger.info("Processing account deletion for %r", email)
+                        logger.info("Processing account delete for %r", email)
                         backend.retire_user(email)
                     elif event_type == "reset":
                         logger.info("Processing account reset for %r", email)
--- a/tokenserver/scripts/process_account_deletions.py
+++ b/tokenserver/scripts/process_account_deletions.py
@@ -63,21 +63,30 @@ def process_account_deletions(config_file, queue_name, aws_region=None,
             if msg is None:
                 continue
             email = None
+            event_type = None
+            generation = None
             # Try very hard not to error out if there's junk in the queue.
             try:
                 body = json.loads(msg.get_body())
                 event = json.loads(body['Message'])
-                # The queue may contain other event types; ignore them.
-                if event["event"] == "delete":
-                    email = event["uid"]
+                event_type = event["event"]
+                email = event["uid"]
+                if event_type == "reset":
+                    generation = event["generation"]
             except (ValueError, KeyError), e:
-                logger.exception("Invalid account-delete message: %s", e)
+                logger.exception("Invalid account message: %s", e)
             else:
-                # Mark the user as retired.
-                # Actual cleanup is done by a separate process.
                 if email is not None:
-                    logger.info("Processing account deletion for %r", email)
-                    backend.retire_user(email)
+                    if event_type == "delete":
+                        # Mark the user as retired.
+                        # Actual cleanup is done by a separate process.
+                        logger.info("Processing account deletion for %r", email)
+                        backend.retire_user(email)
+                    elif event_type == "reset":
+                        logger.info("Processing account reset for %r", email)
+                        user = backend.get_user("sync-1.5", email)
+                        backend.update_user("sync-1.5", user, generation)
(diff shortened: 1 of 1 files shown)


tokenserver/scripts/process_account_deletions.py
```diff
@@ -63,21 +63,30 @@ def process_account_deletions(config_file, queue_name, aws_region=None,
...
-                    backend.retire_user(email)
+                    if event_type == "delete":
+                        # Mark the user as retired.
+                        # Actual cleanup is done by a separate process.
+                        logger.info("Processing account deletion for %r", email)
+                        backend.retire_user(email)
+                    elif event_type == "reset":
+                        logger.info("Processing account reset for %r", email)
+                        user = backend.get_user("sync-1.5", email)
+                        backend.update_user("sync-1.5", user, generation)
```
dannycoates: What's the proper way to get the `service` name here?

rfk: I don't recall there being one since we've never needed to use it like that; we could either make it a magic constant somewhere, or rework this to update the generation number for all services.

rfk:
Good stuff @dannycoates; some high-level comments:

* We should probably rename this file from "process_account_deletions.py" to something more generic, and update the surrounding docstrings, function names, etc.
* I wish we had existing tests for this code; figuring out how to add them may be outside the scope of this PR...
* Watch out for a bad interaction with the code at https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/views.py#L225; it's asserting that the encryption-key hash in client-state does not change unless there's a corresponding change in generation number, to fail early if there's weird client behaviour.  IIUC this patch could cause that to start failing when the clients go to sync again with the updated encryption key.

I think that last point deserves a test of some sort, to ensure that the "sync something, reset password, process reset event, sync something with new key/generation" sequence works as expected.

dannycoates:
> IIUC this patch could cause that to start failing when the clients go to sync again with the updated encryption key

I noticed that section of code but didn't see a problem, perhaps I'm wrong. I assumed if the client has the updated encryption key they'd have the new generation, but those two pieces of info come from different places (`/account/keys` and `/certificate/sign`) so you *could* have one without the other, *sigh*. I don't know if it happens in practice though.

So then what happens in these cases? The error case for updated key without gen seems right from the server perspective. What about the case where gen is updated but not key? Is there a legit reason for this to happen? There's currently no error for this but seems like there should be. I'll look at the client code to see what it does...

This kind of makes me wish we'd included `client_state` in `/sign` and `generation` in `/keys` :shrug:

dannycoates:
OH! What if the `reset` event also included the new `client_state`?

dannycoates:
> What if the reset event also included the new client_state?

nevermind :disappointed: its based off of `kBbytes` not `wrapKb`

rfk:
>  if the client has the updated encryption key they'd have the new generation

This much I think is true in practice.  The problem I'm worried about is with the tokenserver *not* having the new client state, so when the client shows up, the server thinks "you're trying to update client-state but not the generation, wtf?" and errors out.

You could work around this by doing something awfully hacky, like having the reset event increment the stored generation number by one rather than setting it to its final new value.  But, ugh....

> What about the case where gen is updated but not key? Is there a legit reason for this to happen?

IIRC this may happen after a key-preserving password change, rather than a reset.

dannycoates:
> "you're trying to update client-state but not the generation, wtf?"

ok, now I follow you. The behavior I was hoping for was as if the event just acted like a more up-to-date client, but with client_state being based on kB isn't really possible.

> IIRC this may happen after a key-preserving password change, rather than a reset

ah, right. Gen and state are closely tied, but still annoyingly independent.

rfk:
what's the next action on this bug?

tokenserver/scripts/process_account_deletions.py
```diff
@@ -63,21 +63,33 @@ def process_account_deletions(config_file, queue_name, aws_region=None,
...
-                    backend.retire_user(email)
+                    if event_type == "delete":
+                        # Mark the user as retired.
+                        # Actual cleanup is done by a separate process.
+                        logger.info("Processing account delete for %r", email)
+                        backend.retire_user(email)
+                    elif event_type == "reset":
+                        logger.info("Processing account reset for %r", email)
+                        user = backend.get_user("sync-1.5", email)
+                        if generation > user['generation']:
```
rfk: This doesn't need to be conditional, the backend will silently ignore any requests to set the generation number to something smaller than it already is.  Given that, I don't think we even need to worry about additional tests for this feature.

rfk:
Merging this, we can clean up as part of a larger refactor

PR #80, opened 2016-02-11
Rename process_account_deletions script to reflect more generic nature.

rfk:
@mostlygeek you brought it up, you get to review it ;-)

This also adds a warning log on unrecognized event types, but still deletes them from the queue as we're not sending any messages that we can't afford to lose, and this way we don't have to worry about a backlog.


mostlygeek:
R+. Do you think we should repoint v1.2.18 at this change or should we roll v1.2.19?

rfk:
tags are cheap, I'll roll a new one

PR #81, opened 2016-02-24
add fxa uid and device_id to tokens

dannycoates:
This change adds `fxa_uid` and `device_id` to the generated token as well as logging them for metrics.

For `device_id` it relies on the new `fxa-deviceId` claim from FxA coming in v1.57. When not present it will be a hash of `fxa_uid + 'none'`.

### DIFF
--- a/tokenserver/__init__.py
+++ b/tokenserver/__init__.py
@@ -70,7 +70,11 @@ def includeme(config):
 
     # ensure the metrics_id_secret_key is an ascii string.
     id_key = settings.get('fxa.metrics_uid_secret_key')
-    if id_key is not None and isinstance(id_key, unicode):
+    if id_key is None:
+        logger.warning(
+            'fxa.metrics_uid_secret_key is not set. '
+            'This will allow PII to be more easily identified')
+    elif isinstance(id_key, unicode):
         settings['fxa.metrics_uid_secret_key'] = id_key.encode('ascii')
 
     read_endpoints(config)
--- a/tokenserver/views.py
+++ b/tokenserver/views.py
@@ -12,7 +12,7 @@ import tokenlib
 
 from tokenserver.verifiers import get_verifier
 from tokenserver.assignment import INodeAssignment
-from tokenserver.util import json_error, fxa_metrics_uid
+from tokenserver.util import json_error, fxa_metrics_hash
 
 import browserid.errors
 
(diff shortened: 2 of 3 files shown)


rfk:
Is it odd that one has an "fxa" prefix and the other doesn't..?  *shrug*

tokenserver/views.py
```diff
@@ -101,15 +101,29 @@ def valid_assertion(request):
     request.metrics['token.assertion.verify_success'] = 1
     request.validated['assertion'] = assertion
 
-    # log the email with the request
-    email = request.validated['assertion']['email']
+    id_key = request.registry.settings.get("fxa.metrics_uid_secret_key")
+    if id_key is None:
+        id_key = 'insecure'
```
rfk: I think we should at least log a warning here if the key is not configured.

tokenserver/views.py
```diff
@@ -101,15 +101,29 @@ def valid_assertion(request):
...
+    # "legacy" key used by heka active_counts.lua
+    request.metrics['uid'] = fxa_uid_full
     request.metrics['email'] = email
 
-    # Include a unique FxA identifier in the logs, but obfuscate
-    # it for privacy purposes.
-    id_key = request.registry.settings.get("fxa.metrics_uid_secret_key")
-    if id_key:
-        request.metrics['uid'] = fxa_metrics_uid(email, id_key)
+    # "new" keys use shorter values
```
rfk: Is this to keep the size of the token down?  How big is the token now with these extra fields in it?

rfk:
But generally :+1: on tunneling this all the way through

tokenserver/views.py
dannycoates: yes, and also to save some space in redshift. "16 bytes ought to be enough for anybody" :wink:

tokenserver/views.py
dannycoates: Better to log once at startup than every request, yeah?

tokenserver/__init__.py
```diff
@@ -70,7 +70,11 @@ def includeme(config):
 
     # ensure the metrics_id_secret_key is an ascii string.
     id_key = settings.get('fxa.metrics_uid_secret_key')
-    if id_key is not None and isinstance(id_key, unicode):
+    if id_key is None:
+        logger.warning(
+            'fxa.metrics_uid_secret_key is not set. '
+            'This will allow PII to be more easily identified')
```
dannycoates: @rfk better?

rfk: :heart: :+1:

tokenserver/__init__.py
```diff
@@ -70,7 +70,11 @@ def includeme(config):
...
+    elif id_key is not None and isinstance(id_key, unicode):
```
rfk: Although the `id_key is not None` check here is now redundant

rfk:
r+, although I guess you need https://github.com/mozilla-services/tokenserver/pull/83 to get the green tick from CircleCI?

Let's keep an eye out when QA'ing this to just check that the increased token size doesn't break anything.  I can't imagine it will, but there may be e.g. header size limitations that we accidentally bump up against.  And for completeness we'll have to check that the loadtests are accurately reflecting the new token size.

dannycoates:
apparently I don't have merge privileges on this or server-syncstorage

rfk:
you're now a member of "sagrada-devs" which owns these repos; oh man, that's a project code-name I haven't heard in a while...

PR #90, opened 2016-05-13
feat(server): reject assertions if fxa-tokenVerified is false

philbooth:
First stab at implementing `tokenVerified` field validation for FxA-83, sign-in confirmation.

@rfk r?

/cc @vbudhram @shane-tomlinson

### DIFF
--- a/tokenserver/tests/test_service.py
+++ b/tokenserver/tests/test_service.py
@@ -163,6 +163,26 @@ class TestService(unittest.TestCase):
             with self.assertRaises(ValueError):
                 res = self.app.get('/1.0/sync/1.1', headers=headers)
 
+    def test_unverified_token(self):
+        headers = {'Authorization': 'BrowserID %s' % self._getassertion()}
+        # Assertion should not be rejected if fxa-tokenVerified is unset
+        mock_response = {
+            "status": "okay",
+            "email": "test@mozilla.com",
+            "idpClaims": {}
+        }
+        with self.mock_verifier(response=mock_response):
+            self.app.get("/1.0/sync/1.1", headers=headers, status=200)
+        # Assertion should not be rejected if fxa-tokenVerified is True
+        mock_response['idpClaims']['fxa-tokenVerified'] = True
+        with self.mock_verifier(response=mock_response):
+            self.app.get("/1.0/sync/1.1", headers=headers, status=200)
+        # Assertion should be rejected if fxa-tokenVerified is False
+        mock_response['idpClaims']['fxa-tokenVerified'] = False
+        with self.mock_verifier(response=mock_response):
+            res = self.app.get("/1.0/sync/1.1", headers=headers, status=401)
+        self.assertEqual(res.json['status'], 'invalid-credentials')
+
     def test_generation_number_change(self):
         headers = {"Authorization": "BrowserID %s" % self._getassertion()}
         # Start with no generation number.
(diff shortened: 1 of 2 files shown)


tokenserver/views.py
```diff
@@ -96,6 +96,13 @@ def valid_assertion(request):
             raise _unauthorized("invalid-timestamp")
         raise _unauthorized("invalid-credentials")
 
+    try:
+        tokenVerified = assertion['idpClaims']['fxa-tokenVerified']
+    except KeyError:
+        tokenVerified = True
+    if tokenVerified is False:
```
rfk: nit: `if not tokenVerified` would be more idomatic here

tokenserver/views.py
```diff
@@ -96,6 +96,13 @@ def valid_assertion(request):
             raise _unauthorized("invalid-timestamp")
         raise _unauthorized("invalid-credentials")
 
+    try:
+        tokenVerified = assertion['idpClaims']['fxa-tokenVerified']
+    except KeyError:
+        tokenVerified = True
```
rfk: nit: `assertion['idpClaims'].get('fxa-tokenVerified', True)` will have the same effect, and save you the try-except

rfk: It would also be worth adding a comment to explain what's going on here, and why we default to `True`.

tokenserver/tests/test_service.py
```diff
@@ -163,6 +163,19 @@ def test_unauthorized_error_status(self):
...
+            "status": "okay",
+            "email": "test@mozilla.com",
+            "idpClaims": {
+                "fxa-tokenVerified": False
+            }
+        }
+        with self.mock_verifier(response=mock_response):
+            res = self.app.get("/1.0/sync/1.1", headers=headers, status=401)
+        self.assertEqual(res.json['status'], 'invalid-credentials')
+
```
rfk: I think we should also have a test for the explicit `fxa-tokenVerified = True` case.

rfk:
Looks good!  I left some small nits but this is pretty much exactly how I was imagining it would work.

tokenserver/views.py
philbooth: `assertion['idpClaims'].get('fxa-tokenVerified', True)` failed for tests where `idpClaims` isn't set, so I've instead replaced the try/except with a double `get`. Or is there a better way?

Also, the double `get` edged me past the 80th column so I had to introduce a line break. Not sure whether the indentation I went for on my continuation line is okay or visually abhorrent.

philbooth:
Thanks for the feedback @rfk, I've pushed fixes. Let me know if there is a better idiom than the double `get` I used to eliminate the try/except.

tokenserver/views.py
rfk: Oh, good point...the `try/except` may indeed be cleaner overall than a double `.get` then

philbooth:
Have pushed again with the try/except reinstated.

rfk:
Thanks @philbooth, r+ from me; should we just land it, or hold off until we actually land the auth-server side portion of this?

rfk:
Eh, I'm just going to land it so it's easier to e.g. deploy into our dev environment.  We can always tweak in follow-up PRs.

rfk:
@shane-tomlinson @philbooth FYI, I went ahead and deployed this to https://confirm-signin.dev.lcip.org/

PR #93, opened 2016-07-22
Return hashed fxa uid

### LINKED ISSUE #92: Include the hashed fxa_uid in the tokenserver response body.
The server logs a hashed value to identify a specific Sync user, and the Desktop telemetry pings would like to record that same value so we can "join" server-side and client-side metrics to a single user.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1288445#c6 and later for more context - the rough proposal is that the code at https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/views.py#L302 include the "fxa_uid" in the main token body, although possibly with a name that includes "hashed_" so that it isn't easily confused with the fxa uid as returned by the auth server.

rfk:
Fixes #92; @mhammond @dannycoates r?


mhammond:
Looks perfect!

rfk:
merging r=:markh

PR #95, opened 2016-10-19
Reduce docker image size from 776.5MB => 283.7MB

mostlygeek:
- change base to python:2.7-slim
- remove g++ package and deps when using pip

@rfk r? please


rfk:
Nice!

Issue #94, opened 2016-10-19
Add a dockerflow endpoints


mostlygeek:
When the docker container is built by circleci a version.json file is written into '/app/version.json'. When Tokenserver is run we should be able to curl `https://tokenserver_host/__version__` and it should return this JSON document. 

If the version.json file does not exist, a blank `{}` is an acceptable response. 

Also add a `__lbheartbeat__` endpoint that returns a 200 with a blank body.

Natim:
Fixed with https://github.com/mozilla-services/tokenserver/commit/db8089ff1c593b793ef5719358116b980ea8d882

PR #98, opened 2017-02-20
Docker flow endpoints

Natim:
Refs #94

I literally took the same views behaviour are the one we have in kinto.core

r? @leplatrem 

@mostlygeek can you validate that it works when version.json is in the /app/version.json directory?


tokenserver/views.py
```diff
@@ -308,3 +312,46 @@ def return_token(request):
...
+def version_view(request):
+    try:
+        return version_view.__json__
+    except AttributeError:
+        pass
+
+    files = [
+        './version.json',  # Default is current working dir.
+        os.path.join(ORIGIN, 'version.json'),  # Relative to the package root.
+        os.path.join(HERE, 'version.json')  # Relative to this file.
```
leplatrem: this would deserve some documentation I think

tokenserver/tests/test_service.py
```diff
@@ -119,6 +121,21 @@ def test_discovery(self):
...
+        self.app.get('/__version__', status=404)
+
+    def test_version_returns_file_in_current_folder_if_present(self):
+        content = {'version': '0.8.1'}
+        fake_file = mock.mock_open(read_data=json.dumps(content))
+        with mock.patch('os.path.exists'):
+            with mock.patch('tokenserver.views.open', fake_file, create=True):
+                response = self.app.get('/__version__')
+                self.assertEquals(response.json, content)
+
```
leplatrem: nit: I don't know if you want to cover all this but you should also have:
* 404 if file missing
* version.json override from the 3 supported locations
* memoized result

tokenserver/views.py
Natim: Shall I remove this case?

PR #99, opened 2017-02-20
Update working environment.

Natim:
r? @leplatrem


Dockerfile
```diff
@@ -10,7 +10,7 @@ COPY ./requirements.txt /app/requirements.txt
 RUN apt-get -q update \
     && apt-get -q --yes install g++ \
     && pip install --upgrade --no-cache-dir -r requirements.txt \
-         gunicorn nose flake8 \
+         gunicorn nose flake8 mock webtest unittest2 \
```
leplatrem: Just out of curiosity, why do we want dev deps like nose or mock in the Docker instance? (A comment would be valuable IMO)

Makefile
```diff
@@ -1,36 +1,75 @@
...
+
+maintainer-clean: distclean
+	rm -fr local/ .tox/
+
+NAME := tokenserver
+SOURCE := $(shell git config remote.origin.url | sed -e 's|git@|https://|g' | sed -e 's|github.com:|github.com/|g')
+VERSION := $(shell git describe --always --tag)
+COMMIT := $(shell git log --pretty=format:'%H' -n 1)
+version-file:
+	echo '{"name":"$(NAME)","version":"$(VERSION)","source":"$(SOURCE)","commit":"$(COMMIT)"}' > version.json
```
leplatrem: 👍

Dockerfile
leplatrem: And why not use `-r dev-requirements.txt` then ?

dev-requirements.txt
```diff
@@ -0,0 +1,4 @@
+flake8
+nose
+webtest
```
leplatrem: mock?

setup.py
```diff
@@ -4,7 +4,24 @@
...
+requires = [
+    'cornice < 2',
+    'mozsvc >= 0.8',
+    'PyBrowserID',
+    'SQLAlchemy',
+    'boto',
+    'hawkauthlib',
+    'testfixtures',
+    'tokenlib',
+    'umemcache',
```
leplatrem: Out of curiosity, why do we have so many new dependencies?

Dockerfile
Natim: +1 to use dev-requirements.txt now that it exists.
Apparently they want to run tests from the Docker and it is the only place to install the dependencies in it.

setup.py
Natim: They are not new they were only in the requirement.txt file before.

PR #100, opened 2017-02-20
Upgrade to cornice 2 and drop support for Python 2.6

Natim:
Based on #99
r? @mozilla-services/verbatim-project

### DIFF
--- a/tokenserver/tests/support.py
+++ b/tokenserver/tests/support.py
@@ -2,6 +2,9 @@ import os
 
 from browserid.verifiers.local import LocalVerifier
 from browserid.tests.support import (make_assertion, get_keypair)
+from webob import exc
+from webob.dec import wsgify
+from pyramid.httpexceptions import HTTPException
 
 # if unittest2 isn't available, assume that we are python 2.7
 try:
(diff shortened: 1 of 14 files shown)


tokenserver/tests/support.py
```diff
@@ -48,3 +51,17 @@ def get_assertion(email, audience="*", issuer='browserid.org',
...
+        self.app = app
+        if hasattr(app, 'registry'):
+            self.registry = app.registry
+
+    @wsgify
+    def __call__(self, request):
+        try:
+            return request.get_response(self.app)
+        except (exc.HTTPException, HTTPException) as e:
+            return e
```
leplatrem: What happens if we don't use this wrapper? Are there many tests that rely on those exceptions?

Is it because we don't use WebTest that we have to do this?

Natim: To be honest I did not look deeper about that. I just saw that it was imported and no more available in cornice so I backported it.

tokenserver/views.py
```diff
@@ -59,7 +59,7 @@ def _invalid_client_state(reason, **kw):
 
 
 # validators
-def valid_assertion(request):
+def valid_assertion(request, **kwargs):
```
rfk: Is the extra `**kwargs` here necessary to work with latest cornice?

Natim: Yes

PR #106, opened 2017-04-27
Return a distinct error when rejecting a new user with otherwise valid credentials

rfk:
Fixes https://github.com/mozilla-services/syncserver/issues/89 for self-hosters; @Natim r?

I'll also update the API docs if/when we merge this.


rfk:
Thanks!  Docs commit at https://github.com/mozilla-services/docs/commit/7eb4b412d3ab5ec46b280eff312ace32e7cf27e6

PR #107, opened 2017-05-03
Support separate "uid" and "iss" fields in account messages.

rfk:


rfk:
The fact that these messages have an email-like-thing in their "uid" field is a bit of a code smell, and I want to clean it up before we start sending the event stream to more reliers:

  https://github.com/mozilla/fxa-auth-server/pull/1879

Since tokenserver is the only thing that treats FxA uids as if they were an email address, let's put the burden on tokenserver to do the right thing here.  @Natim r?

tokenserver/scripts/process_account_events.py
```diff
@@ -88,7 +88,15 @@ def process_account_event(config, body):
         body = json.loads(body)
         event = json.loads(body['Message'])
         event_type = event["event"]
-        email = event["uid"]
+        uid = event["uid"]
+        try:
+            email = "%s@%s" % (uid, event["iss"])
```
Natim: So if ``event["iss"]`` doesn't exists it means we are using the previous API.

Can we use ``if "iss" in event`` instead or ``if '@' not in uid``?

rfk: Sorry, I'm not totally sure what you're getting at here; do you mean something like:

```
if "iss" in event:
    email = "%s@%s" % (uid, event["iss"])
else:
    email = uid
```

Natim: Yes

rfk:
Thanks @Natim! Merging with your suggested change.

PR #109, opened 2017-06-09
Restore MySQL-related dependencies, add MySQL tests to travis.

rfk:
The dockerflow refactor appears to have accidentally dropped the mysql-related dependencies that we use in production.  This restores them, and tries to ensure that tests are run under MySQL in travis.

I'm not sure how to get the circle tests to run with mysql against the build docker image, @mostlygeek suggestions?


tokenserver/assignment/sqlnode/sql.py
```diff
@@ -608,7 +613,7 @@ def get_best_node(self, service):
                              nodes.c.capacity > nodes.c.current_load,
                              nodes.c.downed == 0)
                 fields = {
-                    'available': sqlfunc.min(
+                    'available': self._sqlfunc_min(
```
rfk: Surprising no-one, running the MySQL tests in CI appears to have uncovered a bug in the MySQL support.  I don't think the previous version of this can ever have worked...perhaps we just don't exercise this code path very often in production?

rfk:
@Natim could you please give this a quick once-over?

mostlygeek:
I tweaked circle.yml so it runs the tests against mysql instead of defaulting to sqlite. Please give that an r? too.

rfk:
@mostlygeek thanks, LGTM!

jrgm:
I branched fix-prod-dependencies to dockerpush-fix-prod-dependencies and added circle logic to push docker images on branches matching /^(master|dockerpush.*)$/. - https://github.com/mozilla-services/tokenserver/commit/f8f3ce12145d80689416cb3c3fb28ba15197c941

With that I tried building the stage pipeline on tag dockerpush-fix-prod-dependencies but I have the error `docker: Error: The 'Paste' distribution was not found and is required by the application`.

rfk:
I'm going to go ahead and merge this so we can get the deploy out, thanks @jrgm @mostlygeek

PR #110, opened 2017-06-09
circleci: push docker images for branches matching /^dockerpush.*$/

jrgm:
This adds the ability to have docker images based on branches.

r? - @mostlygeek, @rfk


jrgm:
For adding to the `fix-prod-dependencies` branch (although no image will be pushed for that branch since the branch name doesn't match).

PR #114, opened 2018-03-26
Fix validation of X-Client-State header.

### LINKED ISSUE #113: Broken validation of X-Client-State header
The validation under https://github.com/mozilla-services/tokenserver/blob/a74a721280fe9bad30d9495a9be614ce61e912a3/tokenserver/views.py#L170 is broken, it will only validate the beginning of a string. So "foobar\n\r\t<>'" will be considered a valid client state. Not that this has any impact on anything from what I can tell, but still a bug.

rfk:
Fixes #113; @vladikoff could you pls give me a quick sanity-check r? here?

### DIFF
--- a/tokenserver/views.py
+++ b/tokenserver/views.py
@@ -167,7 +167,7 @@ def valid_client_state(request, **kwargs):
     """Checks for and validates the X-Client-State header."""
     client_state = request.headers.get('X-Client-State', '')
     if client_state:
-        if not re.match("[a-zA-Z0-9._-]{1,32}", client_state):
+        if not re.match("^[a-zA-Z0-9._-]{1,32}$", client_state):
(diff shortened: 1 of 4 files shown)


tokenserver/views.py
```diff
@@ -167,7 +167,7 @@ def valid_client_state(request, **kwargs):
     """Checks for and validates the X-Client-State header."""
     client_state = request.headers.get('X-Client-State', '')
     if client_state:
-        if not re.match("[a-zA-Z0-9._-]{1,32}", client_state):
+        if not re.match("^[a-zA-Z0-9._-]{1,32}$", client_state):
```
vladikoff: 👍

vladikoff:
@rfk ran some tests, validation looks stronger now 👍

PR #115, opened 2018-04-03
Bug 1455219 - Accept FxA OAuth tokens for authorization.

rfk:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1455219

The main idea here is to fork the "verifier" logic into two different mechanisms - the current one that accepts BrowserID assertions, and a new one that accepts OAuth tokens.  The actual verification of tokens is handled by [PyFxA](https://github.com/mozilla/PyFxA), this just plumbs it in to tokenserver and smooshes the result into a form that matches the output from BrowserID.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,6 @@
+-r requirements.txt
 flake8
 mock
 nose
+responses
 webtest
--- a/tokenserver/tests/test_service.py
+++ b/tokenserver/tests/test_service.py
@@ -84,7 +84,7 @@ class TestService(unittest.TestCase):
                 verifier.__dict__["verify"] = orig_verify_method
 
     def _getassertion(self, **kw):
-        kw.setdefault('email', 'tarek@mozilla.com')
+        kw.setdefault('email', 'test1@example.com')
         kw.setdefault('audience', 'http://tokenserver.services.mozilla.com')
         return make_assertion(**kw).encode('ascii')
 
--- a/tokenserver/tests/test_service.py
+++ b/tokenserver/tests/test_service.py
@@ -15,8 +15,12 @@ from mozsvc.config import load_into_settings
 from mozsvc.plugin import load_and_register
 
 from tokenserver.assignment import INodeAssignment
-from tokenserver.verifiers import get_verifier
+from tokenserver.verifiers import (
+    get_browserid_verifier,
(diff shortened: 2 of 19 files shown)


rfk:
@bbangert would you or someone from your team like to volunteer to be my review-buddy on this python server patch?  It's still WIP but would be good to figure out review situation ahead of time, and any early feedback is always appreciated.

tokenserver/tests/test_service.py
```diff
@@ -88,6 +117,9 @@ def _getassertion(self, **kw):
         kw.setdefault('audience', 'http://tokenserver.services.mozilla.com')
         return make_assertion(**kw).encode('ascii')
 
+    def _gettoken(self, email='tarek@mozilla.com'):
```
jrconlin: maybe "example@mozilla.com" instead? tarek probably gets enough email as is.

tokenserver/verifiers.py
```diff
@@ -144,3 +159,71 @@ def verify(self, assertion, audience=None):
...
+    with an email address, it combines the FxA uid with the hostname of
+    the corresponding FxA BrowserID issuer.  For non-standard FxA hosting
+    setups this might require it to dynamically discover the BrowserID
+    issuer by querying the OAuth verifier's configuration.
+    """
+    implements(IOAuthVerifier)
+
+    def __init__(self, server_url=None, issuer=None, timeout=None,
+                 scope='https://identity.mozilla.com/apps/oldsync'):
+        if timeout is None:
```
jrconlin: Do you have instances where you pass `None` as the value, or is this more to prevent library accidents?

dev-requirements.txt
```diff
@@ -1,4 +1,5 @@
 flake8
```
jrconlin: nit: include `-r requirements.txt` at the top?

(that way you only have to `pip include -r ` one file.

tokenserver/tests/test_service.py
rfk: lol, the tests have a bunch of existing references to tarek but yeah, it's probably time for them to go :-)

tokenserver/verifiers.py
rfk: I'll make it a plain old default arg.

PR #116, opened 2018-05-01
Bug 1457851 - Add back-channel generation update on password change.

rfk:
This will help to more quickly lock out disconnected devices when the user changes their account password, as described in [Bug 1457851 Comment 7](https://bugzilla.mozilla.org/show_bug.cgi?id=1457851#c7).  @jrconlin are you up for another, hopefully simpler r? here?

### DIFF
--- a/tokenserver/scripts/process_account_events.py
+++ b/tokenserver/scripts/process_account_events.py
@@ -6,13 +6,20 @@
 Script to process account-related events from an SQS queue.
 
 This script polls an SQS queue for events indicating activity on an upstream
-account.  The following event types are currently supported:
+account, as documented here:
+
+  https://github.com/mozilla/fxa-auth-server/blob/master/docs/service_notifications.md
+
+The following event types are currently supported:
 
   * "delete":  the account was deleted; we mark their records as retired
                so they'll be cleaned up by our garbage-collection process.
 
   * "reset":   the account password was reset; we update our copy of their
                generation number to disconnect other devices.
+
+  * "passwordChange":  the account password was changed; we update our copy
+                       of their generation number to disconnect other devices.
 
 Note that this is a purely optional administrative task, highly specific to
 Mozilla's internal Firefox-Accounts-supported deployment.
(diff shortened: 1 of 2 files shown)


tokenserver/scripts/process_account_events.py
```diff
@@ -111,23 +118,46 @@ def process_account_event(config, body):
...
+    When the FxA server sends us an update to the user's generation
+    number, we want to update our high-water-mark in the DB in order to
+    immediately lock out disconnected devices.  However, since we don't
+    know the new value of X-Client-State that goes with it, we can't just
+    record the new generation number in the DB.  If we did, the first
+    device that tried to sync with the new generation number would appear
+    to have an incorrect X-Client-State value, and would be rejected.
+
+    Instead, we take advantage of the fact that it's a timestamp, and write
+    it into the DB at one millisecond less than its current value.  This
```
jrconlin: Q: how do you deal with clock skew?

I'm not familiar enough with the code to know where the source of the generation value comes from, but if it comes from a different machine, could that millisecond difference become more significant, or are we expecting the check-ins to be at a far greater scale and thus not a big problem?

rfk: Thanks @jrconlin!

> Q: how do you deal with clock skew?

We don't really do tdo much about that locally here.  This number comes from the core FxA auth server, we assume that FxA manages it as a monotonic integer.  We don't ever compare it to any other timestamps or try to do calculations with it apart from the one here.

The one consideration we do have, is that we only ever write new values to the db when they're actually greater than the current value:

  https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/assignment/sqlnode/sql.py#L65

This helps us ensure it really is monotonic, even if SQS messages are received out-of-order.

> are we expecting the check-ins to be at a far greater scale and thus not a big problem?

Basically this in practice - this number changes when the user does some pretty major action on their account, like resetting their password.  Users who do that more than once per millisecond likely have bigger problems :-)

PR #118, opened 2018-05-09
Bug 1459113 - Error out cleanly when a verifier backend isn't loaded.

rfk:
This ensures that all routes will react cleanly when one of the auth backends is not loaded, which should fix https://bugzilla.mozilla.org/show_bug.cgi?id=1459113#c1


kitcambridge:
Cool, r+! I think this is the first use of `try...except...pass` I've seen in the wild.

PR #117, opened 2018-05-03
Migrate to CircleCI 2

autrilla:
Builds on my fork (https://circleci.com/gh/autrilla/tokenserver/15, disregard docker push errors).

PR #119, opened 2018-05-10
Migrate to CircleCI 2

rfk:
Testing https://github.com/mozilla-services/tokenserver/pull/117 as a feature branch


rfk:
I've updated this based on latest patterns from other repos.  It's now failing the "push to dockerhub" step with:

```
Error response from daemon: Get https://registry-1.docker.io/v2/: unauthorized: incorrect username or password
```

So I guess we need to tweak circle config to use update dockerhub credentials?  @autrilla could you please have a look at your convenience?

autrilla:
The CircleCI job was using Dockerhub credentials that were rotated because they leaked, I've set up good credentials and it built and pushed to Dockerhub successfully.

Issue #121, opened 2018-09-27
Pass the FxA uid and x-client-state through to storage nodes


rfk:
Our new durable storage nodes [1] will be backed by automagically-scalable databases such as Google Spanner, which implement their own internal sharding logic.  For such databases, the way that tokenserver allocates sequential integer userids will be a liability, since sequential primary keys cause hotspots that defeat the internal sharding logic [2].

The FxA uid would be a suitable primary key, since it's a randomly-generated string rather than a sequential integer.  But the storage nodes can't just use the FxA uid as their primary key, because they need to enforce separation of data that belongs to the same user but is encrypted with different keys.

I think the best primary key for such databases will be a compound primary key consisting of the FxA uid and the "x-client-state" token, which is a unique identifier for the current encryption key.  This will allow them to segregate data that has been encrypted with different keys, while still being able to enumerate all records associated with a particular user (which is important for GDPR compliance and for general cleanup purposes).

To allow this to work, tokenserver will need to include the user's FxA uid and x-client-state token in the Hawk token id payload constructed here:

  https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/views.py#L403

As well as in the synthetic tokens that it uses for doing background cleanup here:

  https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/scripts/purge_old_records.py#L91

The storage node can then extract them from the Hawk token and use them as primary key for its database queries.

Some related thoughts:

* This scheme will prevent us from doing a "synthetic" node migration that wipes a user's sync data without resetting their password.  That's probably for the best, as the whole point of building durable storage nodes is that we'll stop doing non-password-reset-related node migrations.

* The Hawk token id already contains a field called "fxa_uid", but this field is the hashed "metrics id" generated at [3].  We shouldn't use that for db primary key because it's not a guaranteed stable identifier (it's produced via hmac with a server-side secret).  We could use a different field name, or we could figure out how to migrate the existing field to a new name like "hashed_fxa_uid".

* When tokenserver was originally designed, we intentionally did not pass through the underlying user identity, and the sequential numeric uids provided a kind of anonymization of the data on the storage nodes.  I don't think anonymization is providing us much value in practice these days, and no other FxA-connected service does it, so I'm not worried about dropping it.  But perhaps we should get a quick sanity-check from legal etc?

[...]

PR #122, opened 2018-09-28
Include raw FxA user-id and key-id in Hawk token.

rfk:
This is a sketch of how we might implement #121, sending the raw FxA uid and key-id as part of the hawk token id.   It will need some tests, and it may be an opportunity to refactor out some common helper functions for this stuff.

Connects to #121

/cc @bbangert @philbooth @pjenvey

### DIFF
--- a/tokenserver/views.py
+++ b/tokenserver/views.py
@@ -102,17 +102,53 @@ def valid_authorization(request, **kwargs):
     name, token = authz
 
     if name.lower() == 'browserid':
-        return _valid_browserid_assertion(request, token)
+        _validate_browserid_assertion(request, token)
+    elif name.lower() == 'bearer':
+        _validate_oauth_token(request, token)
+    else:
+        resp = _unauthorized(description='Unsupported')
+        resp.www_authenticate = ('BrowserID', {})
+        raise resp
 
-    if name.lower() == 'bearer':
-        return _valid_oauth_token(request, token)
+    authorization = request.validated['authorization']
+    email = authorization['email']
+    request.validated['fxa_uid'] = email.split("@", 1)[0]
 
-    resp = _unauthorized(description='Unsupported')
-    resp.www_authenticate = ('BrowserID', {})
-    raise resp
+    # For metrics purposes we expose a "anonymized" version of the
+    # FxA uid which as been hmaced with a server-side secret key.
+    # We call this the "metrics uid" and it correlates to identifiers
+    # used in other systems, such as client-side telemetry.
+    #
+    # For legacy reasons the active_counts.lua script expects a longer
+    # id stored in the key "uid", while other scripts can accept a
+    # shorter id stored in the key "metrics_uid".
+    request.metrics['email'] = email
+    id_key = request.registry.settings.get("fxa.metrics_uid_secret_key")
+    if id_key is None:
+        id_key = 'insecure'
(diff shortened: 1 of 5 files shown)


tokenserver/views.py
```diff
@@ -151,29 +151,31 @@ def _valid_browserid_assertion(request, assertion):
...
+    request.metrics['uid'] = hashed_fxa_uid_full
     request.metrics['email'] = email
 
     # "new" keys use shorter values
-    fxa_uid = fxa_uid_full[:32]
-    request.validated['fxa_uid'] = fxa_uid
-    request.metrics['fxa_uid'] = fxa_uid
+    hashed_fxa_uid = hashed_fxa_uid_full[:32]
+    request.validated['hashed_fxa_uid'] = hashed_fxa_uid
+    request.metrics['hashed_fxa_uid'] = hashed_fxa_uid
```
philbooth: Just bike-shedding this property name for a moment (sorry), but I quite like how your explanatory comment calls this a hashed metrics uid, it just makes the distinction a bit clearer. Would `hashed_metrics_uid` be a more helpful name here?

philbooth: (or even just `metrics_uid` as the hashing seems like an implementation detail)

rfk: Bikeshedding very welcome, although we'll need to take care not to break any consumers of the existing names (e.g. changing the key used for `request.metrics` might not be possible without also updating some downstream lua filters).  I think including "metrics" in the name explicitly makes sense, esp. since the main reason we do it this way is for amplitude.

rfk:
OK, I've cleaned this up by extracting duplicated logic into the main `valid_authorization` function, and I've added some basic tests.  I think it's ready for an r? but **DO NOT MERGE** until we've sanity-checked whether any metrics scripts will break by renaming `request.metrics["fxa_uid"]` to `request.metrics["metrics_uid"]`.

rfk:
Ben, can you please review or delegate?

rfk:
> I assume the random key we'll want to use is the hashed fxa uid + the kid?

I recommend trying to somehow us a compound primary key, either as two columns or as a concatenated string "uid + kid", so that you can enumerate all data stored for a particular uid.

rfk:
(The `uid` will contain all the randomly-distributed-ness that you need for sharding etc)

bbangert:
Cool, Spanner lets us do composite keys, so we can include both of them as the primary key.

rfk:
> DO NOT MERGE until we've sanity-checked whether any metrics scripts will break by
> renaming request.metrics["fxa_uid"] to request.metrics["metrics_uid"].

Mailing list discussion didn't yield any blockers to this renaming; see https://bugzilla.mozilla.org/show_bug.cgi?id=1500830 for some followups if interested.

rfk:
> DO NOT MERGE until we've sanity-checked whether any metrics scripts will break by
> renaming request.metrics["fxa_uid"] to request.metrics["metrics_uid"].

Mailing list discussion didn't yield any blockers to this renaming; see https://bugzilla.mozilla.org/show_bug.cgi?id=1500830 for some followups if interested.

rfk:
I have a follow-up change I want to do to enable more accurate tracking of the "fxa_kid" property, but that depends on some FxA-side changes so I'm going to go ahead and merge this.

rfk:
Issue for the follow-up change: https://github.com/mozilla-services/tokenserver/issues/124

PR #131, opened 2019-02-22
provide `python` target on entrypoint for workers

jrgm:
The worker tasks start with a python command line like "python -m tokenserver.scripts.process_account_events -v ./settings.ini ts-account-change-stage". So `docker-entrypoint.sh` needs a case to handle and exec that command.

(Something about this doesn't reallly feel right, though).

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -23,6 +23,10 @@ case "$1" in
             --log-config "$_SETTINGS_FILE"
         ;;
 
+    python)
+        exec "$@"
+        ;;
+
     test_all)
         $0 test_flake8
         $0 test_nose


jrgm:
> Looks fine, but we could also override ENTRYPOINT I guess?

Yeah. I considered that but I don't see with Puppet docker::run to override entrypoint.

jrgm:
Oops. I really shouldn't have merged this for this repo, but my brain was on autopilot.

jrgm:
Say @rfk - can you prep a 1.4.2 that includes this change?

rfk:
Yep: https://github.com/mozilla-services/tokenserver/releases/tag/1.4.2

Enjoy!

PR #137, opened 2019-04-15
Reduce annoying log messages in stage loadtests

rfk:
This fixes 2 out of 4 causes of annoying log messages in stage tokenserver loadtests that were reported in [Bug 1533257](https://bugzilla.mozilla.org/show_bug.cgi?id=1533257):

* When the oauth server returns `errno=108`, that just means that the OAuth token has expired. We don't need to log a full traceback in this case.
* When purging old records, don't try to delete from the node if we know it's down.

The other 2 causes will have to be fixed elsewhere.

### DIFF
--- a/tokenserver/scripts/purge_old_records.py
+++ b/tokenserver/scripts/purge_old_records.py
@@ -58,11 +58,16 @@ def purge_old_records(config_file, grace_period=-1, max_per_loop=10,
                 }
                 rows = list(backend.get_old_user_records(service, **kwds))
                 for row in rows:
-                    logger.info("Purging uid %s on %s", row.uid, row.node)
-                    if row.node is not None:
+                    # Don't attempt to purge data from downed nodes.
+                    # Instead wait for them to either come back up or to be
+                    # completely removed from service.
+                    if row.node is None:
+                        backend.delete_user_record(service, row.uid)
+                    elif not row.downed:
+                        logger.info("Purging uid %s on %s", row.uid, row.node)
                         delete_service_data(config, service, row,
                                             timeout=request_timeout)
-                    backend.delete_user_record(service, row.uid)
+                        backend.delete_user_record(service, row.uid)
                 if len(rows) < max_per_loop:
                     break
     except Exception:
--- a/tokenserver/tests/test_purge_old_records.py
+++ b/tokenserver/tests/test_purge_old_records.py
@@ -26,19 +26,21 @@ class TestPurgeOldRecordsScript(unittest.TestCase):
     so we can test that data-deletion requests go through ok.
     """
 
-    def setUp(self):
-        super(TestPurgeOldRecordsScript, self).setUp()
-
+    @classmethod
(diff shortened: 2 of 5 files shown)


tokenserver/tests/test_purge_old_records.py
```diff
@@ -26,19 +26,21 @@ class TestPurgeOldRecordsScript(unittest.TestCase):
...
-        self.service = make_server("localhost", 8002, self._service_app)
-        target = self.service.serve_forever
-        self.service_thread = threading.Thread(target=target)
-        self.service_thread.start()
+        cls.service_requests = []
+        cls.service_node = "http://localhost:8002"
+        cls.service = make_server("localhost", 8002, cls._service_app)
+        target = cls.service.serve_forever
+        cls.service_thread = threading.Thread(target=target)
+        cls.service_thread.start()
```
rfk: This test suite spins up a localhost server on a background thread. Previously it was only used for a single test, and adding a second test caused it to start blocking and timing out, for reasons I don't really understand (maybe something not done quite right in tearing down the server?).  Anyway, rather than pull my hair out trying to track down the cause, I opted to re-use the instance across test runs, because it's trivial to clear its internal state.

tokenserver/scripts/purge_old_records.py
```diff
@@ -58,11 +58,17 @@ def purge_old_records(config_file, grace_period=-1, max_per_loop=10,
...
-                                            timeout=request_timeout)
-                    backend.delete_user_record(service, row.uid)
+                    # Don't attempt to purge data from downed nodes.
+                    # Instead wait for them to either come back up or to be
+                    # compeltely removed from service.
+                    if row.node is None:
+                        backend.delete_user_record(service, row.uid)
+                    elif not row.downed:
+                        logger.info("Purging uid %s on %s", row.uid, row.node)
+                        if row.node is not None:
```
pjenvey: nit: the elif confirms row.node is not None, this check isn't necessary

tokenserver/scripts/purge_old_records.py
```diff
@@ -58,11 +58,17 @@ def purge_old_records(config_file, grace_period=-1, max_per_loop=10,
...
                 rows = list(backend.get_old_user_records(service, **kwds))
                 for row in rows:
-                    logger.info("Purging uid %s on %s", row.uid, row.node)
-                    if row.node is not None:
-                        delete_service_data(config, service, row,
-                                            timeout=request_timeout)
-                    backend.delete_user_record(service, row.uid)
+                    # Don't attempt to purge data from downed nodes.
+                    # Instead wait for them to either come back up or to be
+                    # compeltely removed from service.
```
mhammond: just saw this in my inbox - s/compeltelycompletely/

tokenserver/scripts/purge_old_records.py
rfk: thanks, nice catch, this is left over from a previous factoring

PR #140, opened 2019-05-21
Bug 1553020 - Don't mock `oauth.default_issuer` for loadtests.

rfk:
Ref https://bugzilla.mozilla.org/show_bug.cgi?id=1553020

Mocking `oauth.default_issuer` means that live users can't authenticate
properly via OAuth, because we'll treat them as "uid@mock-issuer" rather
than "uid@live-issuer", preventing them from accessing their sync data.

Instead, ensure that all mock OAuth tokens are explicitly marked as coming
from the mock issuer, and let the default issuer be read from live config
on the target OAuth server.

### DIFF
--- a/loadtest/mock-oauth-cfn.yml
+++ b/loadtest/mock-oauth-cfn.yml
@@ -3,12 +3,9 @@
 # service that can mock out FxA OAuth verification responses.
 #
 # When deployed, this API will proxy all HTTP requests through to a live
-# FxA OAuth server, with the following exceptions:
-#
-#  * `GET /config` will return a mocked default issuer.
-#  * `POST /v1/verify` will attempt to parse the submitted token as JSON.
-#     If it succeeds, then it will use the `status` and `body` fields from
-#     that JSON to return a mocked response.
+# FxA OAuth server except that `POST /v1/verify` will attempt to parse
+# the submitted token as JSON. If it succeeds, then it will use the `status`
+# and `body` fields from that JSON to return a mocked response.
 #
 # The idea is to let this API be used as a stand-in for the real FxA OAuth
 # server, and have it function correctly for manual testing with real accounts,
@@ -51,7 +48,7 @@ Parameters:
   MockIssuer:
     Type: "String"
     Default: "mockmyid.s3-us-west-2.amazonaws.com"
-    Description: "The mock BrowserID issuer to advertise in config"
+    Description: "The issuer domain to use for mock tokens"
   DomainName:
     Type: "String"
     Default: "mock-oauth-stage.dev.lcip.org"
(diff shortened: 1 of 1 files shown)


loadtest/mock-oauth-cfn.yml
```diff
@@ -119,28 +119,23 @@ Resources:
...
-              });
-            },
-
             'POST:/v1/verify': function(event, context, callback) {
               try {
                 const token = JSON.parse(event.body).token;
                 const mockResponse = JSON.parse(token);
+                const mockStatus = mockResponse.status || 200;
+                const mockBody = mockResponse.body || {};
+                // Ensure successful responses always come from the mock issuer.
```
linacambridge: Hmm, I don't understand how this comment relates to the block below...are we only including `issuer` in the body because, for failed responses, their issuer doesn't matter?

loadtest/mock-oauth-cfn.yml
```diff
@@ -119,28 +119,23 @@ Resources:
...
+                if (mockStatus < 400) {
+                  mockBody.issuer = "${MockIssuer}";
```
linacambridge: Does anything rely on the old nested format?

loadtest/mock-oauth-cfn.yml
```diff
@@ -119,28 +119,23 @@ Resources:
           }
 
           const HANDLERS = {
-
-            'GET:/config': function(event, context, callback) {
```
linacambridge: Should we update the comment at the top that references `GET /config`?

rfk: yes, good catch, thanks

loadtest/mock-oauth-cfn.yml
rfk: Sorry, not sure what you mean by "the old nested format" here.

loadtest/mock-oauth-cfn.yml
rfk: I've reworded to try to make it clearer.

loadtest/mock-oauth-cfn.yml
linacambridge: Oh, sorry, I meant how the old response was `{ "browserid": { "issuer": "${MockIssuer}" } }'`, but the new response sets `mockBody.issuer`. Does it need to be `mockBody.browserid.issuer`?

rfk: >  old response was `{ "browserid": { "issuer": "${MockIssuer}" } }`

That was for the `/config` endpoint, which we're no longer modifying. The new code is adding an explicit `issuer` field to the `/v1/verify` response (which previously would have been provided implicitly by the `mockBody`, if it had such a field).

linacambridge: Derp. I...don’t know how I missed that it was for a completely different endpoint. :scream: Thanks!

Issue #124, opened 2018-10-23
 Explicitly track the key-rotation timestamp for the user's sync key


rfk:
Over in https://github.com/mozilla/fxa-auth-server/issues/2674 we're going to add a new field to the metadata returned in FxA identity assertions telling us the timestamp at which the user's key material last changed.  This is similar to, but not the same as, the "generation number" that we currently track.

Let's update tokenserver to track both "fxa-generation" and "fxa-keysChangedAt" for each user, so that it can accurately construct the kid for the user's sync key.  This will in turn allow the new durable sync storage server to more accurately purge data that was encrypted with an older key.

Concretely, we currently communicate `fxa_uid` and `fxa_kid` fields to the storage nodes like this:

```
{
    'fxa_uid': user['fxa_uid'],        // e.g. "00001111222233334444555566667788"
    'fxa_kid': user['client-state'],   // e.g. "AABBCCDDEEFF11223344556677889900"
}
```

Given two `fxa_kid` values, it's not possible to tell by inspection which is the "current" key-id and which is out-of-date.   When we hand out [scoped keys for OAuth clients](https://github.com/mozilla/fxa-crypto-relier/blob/34bb99f40ab97d7f0f564abcbfab54fd326adb1d/src/deriver/ScopedKeys.js#L137), we identify the keys by a combination of timestamp and hash, which means they sort lexicographically in a useful fashion.  I think we should do the same here, so we end up sending:

```
{
    'fxa_uid': user['fxa_uid'],                                        // e.g. "00001111222233334444555566667788"
    'fxa_kid': user['keys-changed-at'] + "-" + user['client-state'],   // e.g. "1569299946203-AABBCCDDEEFF11223344556677889900"
}
```

This will allow the storage nodes to easily detect and purge data stored under old keys.

The tricky part will be doing this accurately in the presence of incomplete information, because we haven't been tracking the key-rotation timestamp for existing users.  I believe we need to do the following, in order:

1. For OAuth sync clients, stop treating the key rotation timestamp as being the same as the `generation` number. Otherwise we risk locking OAuth clients out of sync once we start tracking the key-rotation timestamp properly.  I'm working on a PR for this.
2. Get the [FxA changes to track key-rotation timestamp](https://github.com/mozilla/fxa/pull/2570) into production on the FxA side, which will make a new `keysChangedAt` field available to the tokenserver.
[...]

PR #142, opened 2019-09-24
Allow OAuth clients to not report a generation number.

rfk:
For OAuth clients, the timestamp component of the `X-KeyID` header is supposed to represent the time at which the keys were last changed.  We have instead been using is the "generation number", which is the timestamp at which the user's *password* was last changed.  Once FxA starts [separately tracking the key-rotation timestamp](https://github.com/mozilla/fxa/pull/2570), these two numbers will sometimes diverge and we risk locking out OAuth clients if we continue to use it as the generation number.

This change exempts OAuth clients from the generation-number checking logic, by having them always report a generation number of `0`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,8 @@ The `generation` column is used to detect when the user's FxA credentials have b
 and to lock out clients that have not been updated with the latest credentials.
 Tokenserver tracks the highest value of `generation` that it has ever seen for a user,
 and rejects BrowserID assertions in which the `generation` number is less than that high-water mark.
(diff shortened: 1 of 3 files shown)


rfk:
(This is step (1) of the approach outlined in https://github.com/mozilla-services/tokenserver/issues/124)

pjenvey:
I've added the migration and db bits here:

https://github.com/mozilla-services/tokenserver/compare/track-key-rotation-timestamp...track-key-rotation-timestamp2

It sounds like we want fxa content server's related changes deployed *before* this would be?

PR #143, opened 2019-09-24
Track key-rotation timestamp if provided by the IdP

rfk:
This is a quick sketch of the changes that might be required for https://github.com/mozilla-services/tokenserver/issues/124.  It's missing the underlying database migration and any tests, but I wanted to put it up as a kind of explainer for what might be required, and for early feedback.

### DIFF
--- a/tokenserver/tests/test_service.py
+++ b/tokenserver/tests/test_service.py
@@ -473,14 +473,18 @@ class TestService(unittest.TestCase):
             res = self.app.get('/1.0/sync/1.1', headers=headers)
         token = self.unsafelyParseToken(res.json["id"])
         self.assertEqual(token["fxa_kid"], "0000000002345-YmJi")
-        # We'll error if it changes without a change in generation.
-        mock_response["idpClaims"]["fxa-keysChangedAt"] = 4567
-        headers["X-Client-State"] = "636363"
-        with self.mock_browserid_verifier(response=mock_response):
-            res = self.app.get('/1.0/sync/1.1', headers=headers, status=401)
-        self.assertEqual(res.json["status"], "invalid-keysChangedAt")
+        # TODO: ideally we will error if keysChangedAt changes without a
+        # change in generation, but we can't do that until all servers
+        # are running the latest version of the code.
+        # mock_response["idpClaims"]["fxa-keysChangedAt"] = 4567
+        # headers["X-Client-State"] = "636363"
+        # with self.mock_browserid_verifier(response=mock_response):
+        #     res = self.app.get('/1.0/sync/1.1', headers=headers, status=401)
+        # self.assertEqual(res.json["status"], "invalid-keysChangedAt")
         # But accept further updates if both values change in unison.
         mock_response["idpClaims"]["fxa-generation"] = 4567
+        mock_response["idpClaims"]["fxa-keysChangedAt"] = 4567
+        headers["X-Client-State"] = "636363"
         with self.mock_browserid_verifier(response=mock_response):
             res = self.app.get('/1.0/sync/1.1', headers=headers)
         token = self.unsafelyParseToken(res.json["id"])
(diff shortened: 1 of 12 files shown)


pjenvey:
I've added the migration, db bits and some notes here:

https://github.com/mozilla-services/tokenserver/compare/track-key-rotation-timestamp...track-key-rotation-timestamp2

It sounds like we want fxa content server's related changes deployed **before** this PR's change to tokenserver would be?

My question about the new failure in `test_credentials_from_oauth_and_browserid`: will content server begin reporting a `keysChangedAt` for browserId assertions (and only if was aware of one, since it could be NULL)?

rfk:
> It sounds like we want fxa content server's related changes deployed before this PR's change
> to tokenserver would be?

Right, exactly. I think it would need to be staged along the lines of:

1. Ship the tokenserver prep-work from https://github.com/mozilla-services/tokenserver/pull/142, which will help us cope with any rollbacks on the FxA side.
2. Get the FxA-side changes shipped and stable.
3. Ship tokenserver change to actually start tracking and using the `keysChangedAt` timestamp.

> will content server begin reporting a keysChangedAt for browserId assertions 

Yes, and I don't think the value will ever need to be null, because it can safely fall back to using the `verifierSetAt` timestamp if it doesn't have a  `keysChangedAt` timestamp for that user.

rfk:
> Yes, and I don't think the value will ever need to be null, because it can safely fall back
> to using the verifierSetAt timestamp if it doesn't have a keysChangedAt timestamp for that user.

The one caveat here is that if we roll back to an earlier version of FxA that doesn't send this new field, then we'll start receiving assertions without it.

rfk:
Thanks @pjenvey! I pulled in your commit and had a go at updating the tests to show the behaviour I expect from the pending FxA change.

rfk:
OK, I'm feeling pretty happy with this.  I've added some additional testcases, and tried to walk through that activity might look like for a user as we roll out this change both here and in FxA. The tests really highlight the need to have the FxA changes in prod and stable before deploying this to tokenserver, because having FxA start sending `keys_changed_at` would make it hard for us to forward a consistent `fxa_kid` value down to the storage nodes.

I'll work on picking up that thread on the FxA side this week.

@pjenvey what do you think?

tokenserver/views.py
```diff
@@ -364,22 +397,32 @@ def return_token(request):
...
-    # client may have raced with a concurrent update.
+    # Error out if this client provided a generation number, but it is behind
+    # the generation number of some previously-seen client.
     if generation > 0 and user['generation'] > generation:
         raise _unauthorized("invalid-generation")
 
+    # Error out if we previously saw a keys_changed_at for this user, but they
+    # haven't provided one or it's earlier than previously seen. This means
+    # that once the IdP starts sending keys_changed_at, we'll error out if it
+    # stops (becauwe we can't generate a proper `fxa_kid` in this case).
```
pjenvey:
```suggestion
    # stops (because we can't generate a proper `fxa_kid` in this case).
```

tokenserver/tests/test_service.py
```diff
@@ -415,15 +424,122 @@ def test_client_state_change(self):
...
+        headers["X-Client-State"] = "626262"
+        with self.mock_browserid_verifier(response=mock_response):
+            res = self.app.get('/1.0/sync/1.1', headers=headers)
+        token = self.unsafelyParseToken(res.json["id"])
+        self.assertEqual(token["fxa_kid"], "0000000002345-YmJi")
+        # If we roll back the FxA stack so it stops reporting keys_changed_at,
+        # users will get locked out because we can't produce `fxa_kid`.
+        del mock_response["idpClaims"]["fxa-keysChangedAt"]
+        with self.mock_browserid_verifier(response=mock_response):
+            self.app.get('/1.0/sync/1.1', headers=headers, status=401)
```
pjenvey: Might be worth adding `res.json['status']` checks for these new status=401 tests

rfk: good idea; updated

rfk:
Copying @jrgm's question from Bug 1591312 to ask it about this change as well: will it be safe to simultaneously have some tokenserver machines running this new version of the code and some running the previous code?

The basic cases seem OK:

* A user might talk first to a server with old code, then to a server with new code. This is the same as what would happen even if we could do an instantaneous rollout, so it'll be fine - the new code will start tracking `keysChangedAt` as expected.
* A user migth talk to a server with the old code after talking to a server with the new code. In this case they may have a value of `keysChangedAt` recorded in the db. The old code will ignore the new column, allowing the user to access sync successfully.

But I think we could get into trouble with some of the consistency checks in this scenario. It's possible that a server running the old code will update the `client_state` and `generation` fields without persitsing the updated value for `keys_changed_at`, leading to a scenario like:

* We do a partial deployment, leaving some servers running old code and some running new code.
* The user syncs, hits the new server code, and gets `keys_changed_at` written into the db.
* The user does a password reset, changing their key-rotation timestamp.
* The user syncs again, this time hitting the old server code. It writes the new `generation` and `client_state` fields into the db, but ignores the new `keys_changed_at`.
* The user syncs again, this time hitting new server code. It detects that their `keys_changed_at` value has increased, but that `generation` matches what is currently stored in the db. Thinking that the client has a consistency issue with its credentials, it errors out and the user is unable to sync.

I'm going to push a small update to remove that consistency check, since we can't apply it sensibly during the deployment. We can bring it back once the whole system is updates to the new code.

rfk:
@pjenvey could you please check the change in 2c6b26fd66fa7fb4c6d1469ac1b95d07216b0df1 in light of [the comment above](https://github.com/mozilla-services/tokenserver/pull/143#issuecomment-550596209)?

rfk:
The FxA-side change to [report `keysChangdAt`](https://github.com/mozilla/fxa/pull/3060) is now live in production. There's a follow-up FxA change to [make that value more accurate](https://github.com/mozilla/fxa/pull/2570) but I believe we can safely merge and deploy this tokenserver change in parallel with that one.

Making this "ready for review", and re-requesting review due to the above changes related to gradual rollout.

rfk:
> Copying @jrgm's question from Bug 1591312 to ask it about this change as well

Turns out he asked me this in slack *about*  Bug 1591312; putting here for full context:
```
The deploy will put out one canary, and I can run that for a while.
But is “dual version mode” not a good way to run for an extended period with this release?
```

tokenserver/views.py
```diff
@@ -349,12 +374,22 @@ def return_token(request):
...
     updates = {}
     if generation > user['generation']:
         updates['generation'] = generation
+    if keys_changed_at > user['keys_changed_at']:
+        # TODO: If there's a generation number available, then a change
+        # in keys should correspond to a change in generation number.
+        # We can't accurately detect this case until all servers are reliably
+        # writing keys_changed_at to the database, so it will need to wait
+        # and be deployed separately.
+        # if generation > 0 and 'generation' not in updates:
```
pjenvey: This looks sound, my only thought is if this is in fact the only valid case:

>     * It detects that their `keys_changed_at` value has increased, but that `generation` matches what is currently stored in the db.

then could this be stricter?

```suggestion
        if generation > 0 and 'generation' not in updates and generation != user['generation']:
```

but with this being a hopefully temporary situation, removing this check entirely is fine.

rfk: Good point, yes, I think you're right!

rfk: In trying to add a test for this, I realized such auth attempts will already be rejected by the "Error out if this client provided a generation number, but it is behind that of some previously-seen client" logic further down the file. I'm going to keep it because it'll be some nice extra check against a misbehaving FxA server.

PR #149, opened 2019-10-24
Report `node_type` for inclusion in client-side metrics.

rfk:
Ref [Bug https://bugzilla.mozilla.org/show_bug.cgi?id=1582317](https://bugzilla.mozilla.org/show_bug.cgi?id=1582317); obviously extremely WIP!

The idea here is to have tokenserver return a short string value for clients to include in the sync ping, which we can use for segmented analysis of sync client metrics. By putting the server in charge of this value we gain a lot of operational flexibility without having to update client code - for example, if the initial rollout of a `"new`" node type is a bit of a bust, we might like to be able to introduce a `"new2"` node type without disrupting the users already allocated to the previous one.

There are many wayswe could have tokenserver determine the appropriate value to send - we could make it a column in the `nodes` table, we could do a simple string match or regex, or maybe something else. I don't have strong opinions on that front.

### DIFF
--- a/tokenserver/__init__.py
+++ b/tokenserver/__init__.py
@@ -4,6 +4,8 @@
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import sys
+import re
+import fnmatch
 from tokenserver.util import monkey_patch_gevent
 
 
@@ -44,6 +46,9 @@ def includeme(config):
     # initialize the oauth backend if it exists
     if "oauth.backend" in settings:
         load_and_register("oauth", config)
+
+    # initialize node-type classifier
+    load_node_type_classifier(config)
 
     # load apps and set them up back in the setting
     key = 'tokenserver.applications'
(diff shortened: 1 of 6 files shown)


rfk:
Ref https://bugzilla.mozilla.org/show_bug.cgi?id=1582317#c18, doesn't sound like we have any appetite for this approach; closing.

rfk:
Ref https://bugzilla.mozilla.org/show_bug.cgi?id=1582317#c25, let's revisit this :-)

rfk:
Alrighty, here's my take on having tokenserver return a string `node_type` field as part of its response, using a simple wildcard-driven config file entry which would look like this:

```
[tokenserver]
node_type_patterns =
  spanner:*.gcpwhatever.services.mozilla.com
  mysql: *.awsthingy.services.mozilla.com
```

rfk:
Thanks!

> At least 1 approving review is required by reviewers with write access. 

Hrm, @fzzzy doesn't have write access to this repo. We should fix this, but also, @mhammond could you please rubber-stamp this for merge based on review above?

PR #156, opened 2019-11-15
Adjust schemas.py to reflect state of DB after migration scripts.

rfk:
Commit 71a590f85a27eff6ee4cddfe25aaeb33d210baa2 appears to have introduced a subtle difference between the schema declared in `schemas.py` and the one that would be produced by running all our migrations in order.  It added a migration making `nodeid` non-nullable and adding an index on `nodeid`, but updated the schema to make `nodeid` nullable and add an index on the
pair (`service`, `nodeid`).

This commit fixes the schema to match what's in the database in production. For good measure I have also re-ordered the columns to match the order they physically appear in the database in production.  With this change, a db re-created from scratch using the declarations in `schemas.py` should end up being the same shape as in prod.

### DIFF
--- a/tokenserver/assignment/sqlnode/schemas.py
+++ b/tokenserver/assignment/sqlnode/schemas.py
@@ -44,12 +44,12 @@ class _UsersBase(object):
                  nullable=False)
     service = Column(Integer(), nullable=False)
     email = Column(String(255), nullable=False)
-    nodeid = Column(BigInteger(), nullable=True)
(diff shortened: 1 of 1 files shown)


rfk:
(A future improvement here would be to *always* create databases by running the migrations forward, rather than having two sources of truth about the state of the db. For tokenserver we would have to synthesize an initial migration that creates the tables in their original form, trying to `alembic upgrade` on an empty db will currently complain about trying to alter a non-existent table.)

PR #157, opened 2019-11-18
Correctly write keys_changed_at when it's the only updated field.

rfk:
In the SQL backend, when updating a user's details, we try to do an in-place update rather than a replacement when the generation number is the only field that has changed.

Due to a bug, this branch was being taken even if the `keys_changed_at` field has changed, with the result being that the new `keys_changed_at` value just gets dropped on the floor.  This showed up in QA of the v1.5.0 release, when I tried to test against a user that already exists in the db.

The first commit here fixes the above bug, and is pretty trivial.

Unfortunately, I wasn't able to reproduce the bug in tests here, because most of the service API tests are only run against the memory backend, which didn't have the bug! I've therefore added a second commit here, which tries to ensure that these also get run against the SQL backend.  This would have caught the bug in CI if we'd had it previously.

To make this work, I had to modify `test_sql.ini` to more closely match `test_memorynode.ini`, which meant I had to tweak the details in several unrelated tests. So there's a surprisingly amount of churn here sorry, but it's only in incidental details like the names of services used in different tests.

### DIFF
--- a/tokenserver/assignment/sqlnode/sql.py
+++ b/tokenserver/assignment/sqlnode/sql.py
@@ -324,7 +324,7 @@ class SQLNodeAssignment(object):
 
     def update_user(self, service, user, generation=None, client_state=None,
                     keys_changed_at=None, node=None):
-        if client_state is None and node is None:
+        if client_state is None and node is None and keys_changed_at is None:
             # We're just updating the generation, re-use the existing record.
             if generation is not None:
                 params = {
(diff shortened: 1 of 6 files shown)


tokenserver/assignment/sqlnode/sql.py
```diff
@@ -324,7 +324,7 @@ def allocate_user(self, service, email, generation=0, client_state='',
 
     def update_user(self, service, user, generation=None, client_state=None,
                     keys_changed_at=None, node=None):
-        if client_state is None and node is None:
+        if client_state is None and node is None and keys_changed_at is None:
```
rfk: This fixes the actual bug in question.

tokenserver/tests/test_service.py
```diff
@@ -190,10 +191,20 @@ def test_discovery(self):
         })
 
     def test_version_returns_404_by_default(self):
+        # clear cache
+        try:
+            del tokenserver.views.version_view.__json__
+        except AttributeError:
+            pass
```
rfk: These tests started failing now that we run them twice, thanks to caching. Bleh.

PR #160, opened 2019-12-05
Don't force a node-reassigment when recording a new value of keys_changed_at

rfk:
During the initial rollout of keys_changed_at, we may be in a situation where the tokenserver first sees a new value for keys_changed_at after some older version of the server has already processed the node-reassignment for the corresponding change in client_state. We don't want to force users through
another node-reassignment in that case, since they would already have done one.

This changes the sql backend to only force a node-reassignment when client_state changes, and adds some tests to ensure it. The memory backend was accidentally doing the right thing already, although the tests uncovered an issue with handing out mutable references to its internal state that I've had to fix as part of this PR.

### DIFF
--- a/tokenserver/assignment/sqlnode/sql.py
+++ b/tokenserver/assignment/sqlnode/sql.py
@@ -64,15 +64,22 @@ values
      :timestamp, NULL)
 """)
 
-
-_UPDATE_GENERATION_NUMBER = sqltext("""\
+# The `where` clause on this statement is designed as an extra layer of protection,
+# to ensure that concurrent updates don't accidentally move timestamp fields
+# backwards in time. The handling of `keys_changed_at` is additionally weird because
+# we want to treat the defalut `NULL` value as zero.
+_UPDATE_USER_RECORD_IN_PLACE = sqltext("""\
 update
     users
 set
-    generation = :generation
+    generation = COALESCE(:generation, generation),
+    keys_changed_at = COALESCE(:keys_changed_at, keys_changed_at)
 where
     service = :service and email = :email and
-    generation < :generation and replaced_at is null
+    generation <= COALESCE(:generation, generation) and
+    COALESCE(keys_changed_at, 0) <=
+        COALESCE(:keys_changed_at, keys_changed_at, 0) and
+    replaced_at is null
 """)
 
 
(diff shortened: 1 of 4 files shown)


jrgm:
Say @rfk - if you push up a `dockerpush.release-1.5.1-avoid-node-reassignment` branch I can deploy that to stage now for you to confirm this change on stage.

rfk:
👍 pushing now under that name

jrgm:
Stage is now commit  f5c32e69f
```
$  curl -s https://token.stage.mozaws.net/__version__ | python -mjson.tool
{
    "build": "https://circleci.com/gh/mozilla-services/tokenserver/295",
    "commit": "f5c32e69fb69f7bbc23167ce63263b486ff11d9d",
    "source": "https://github.com/mozilla-services/tokenserver",
    "version": ""
}
```

rfk:
Thanks! To help with testing, whenever you have a chance, could you please `UPDATE users SET keys_changed_at = NULL WHERE uid = 5469307` on stage? That's my current user account, and will help me test upgrading from an existing record without the keys-changed-at value set.

jrgm:
> could you please UPDATE users SET keys_changed_at = NULL WHERE uid = 5469307 on stage? 

Done.
```
> select keys_changed_at from users where uid = 5469307;
+-----------------+
| keys_changed_at |
+-----------------+
|            NULL |
+-----------------+
```

rfk:
Thanks! I just ran my test again, did it set `keys_changed_at` for this `uid`? (as opposed to the previous behaviour, where it would have created a new `uid` and marked the existing one as replaced).

jrgm:
```
> select keys_changed_at from users where uid = 5469307;
+-----------------+
| keys_changed_at |
+-----------------+
|   1575525631477 |
+-----------------+
```

rfk:
Awesome, looks like this is working as intended 👍

jrgm:
So, when there's a tag for this should I try this in production tomorrow morning PST, or wait for your morning? (Note: I have an appointment tomorrow that I cannot reschedule again at 2300Z)

jrgm:
but I'd really like to get this done.

rfk:
I don't want to send this prod without an r? from @pjenvey, but perhaps you can coordinate with him for a tag tomorrow morning. I'm happy that it's behaving as I expect.

tokenserver/assignment/sqlnode/sql.py
```diff
@@ -65,14 +65,18 @@
...
 update
     users
 set
-    generation = :generation
+    generation = COALESCE(:generation, generation),
+    keys_changed_at = COALESCE(:keys_changed_at, keys_changed_at)
 where
     service = :service and email = :email and
-    generation < :generation and replaced_at is null
+    generation <= COALESCE(:generation, generation) and
```
pjenvey: I had to stare at these bits for a while. Keeping this logic in Python (probably around an actual sqlalchemy update expression) would help readability, but I'm fine committing it as is for now.

rfk: This is mostly belt-and-braces stuff to ensure that we don't move any timestamps backwards, I'll add more of a comment in the statement in the meantime.

PR #163, opened 2019-12-19
feat: Add percentage routing and migration status to user table

### LINKED ISSUE #159: Adjust routing logic for new users to allow us to point X% of them to Spanner.
We're planning to rollout Durable Sync for brand new users of Sync only to start with. Before we can move to [phase 2](https://docs.google.com/document/d/1ULkyVWBeIt0zYo4AmF5aUI1kRDKCPQQaXVZv1w1oj_M/edit#), we'll need to adjust the logic here to make it easy to route users by a percentage.

Ideally, that percentage can be pulled from a config value somewhere so we can easily modify it without a deploy.

Once a user has been routed to Spanner, we'll need to continue routing them over to Spanner for subsequent sessions.

It looks like the logic for routing a new user lives [here](https://github.com/mozilla-services/tokenserver/blob/eeb982fafcdac187061c7b7b62d92a28e75f0d63/tokenserver/views.py#L371) for reference.

jrconlin:
This adds a few entries into the .ini:

```
[tokenserver]
spanner_entry  = # Name of the spanner node name e.g.  https://spanner.example.com
spanner_node_id = # default spanner node id
migrate_new_user_percentage = # percentage of users to send to spanner

```
*note* the "percentage" is a terrible hack that just sends the first _n_
of every 100 users to spanner.

Closes #159


tokenserver/assignment/sqlnode/sql.py
```diff
@@ -181,6 +181,19 @@
...
+_MIGRATE_USER = sqltext("""\
+update
+    users
+set
+    migration_state = :migration_state,
+    nodeid = :nodeid
+where
+    email = :email
+    and service = :service
+""")
```
rfk: This is going to have unintended side-effects by affecting historical node-assignment records.

The existing queries in this file affect a node migration by marking the user's current record as "replaced" and then creating a new record for the new node assignment. There is a separate background process that goes through and cleans up all the replaced records, including messaging out to the old storage nodes to make sure they delete obsoleted user data.

This new query would change the nodeid on all those historical records, meaning that background process can't do its job properly, and potentially putting us in violation of GDPR and other privacy legislation (because we can no longer delete the user's historical sync data in response to an explicit deletion request).

Would is suffice to follow the same approach here, adding a new recording and marking the old ones as replaced, without modifying the historical records?

tokenserver/assignment/sqlnode/migrations/versions/8440ac37978a_migration.py
```diff
@@ -0,0 +1,23 @@
...
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = '8440ac37978a'
+down_revision = '75e8ca84b0bc'
+
+
+def upgrade():
+    op.add_column("users", sa.Column("migration_state", sa.String))
```
rfk: If you're adding columns here, please update the schema description in the README to match.

tokenserver/assignment/sqlnode/sql.py
```diff
@@ -647,6 +688,8 @@ def get_best_node(self, service):
                 res.close()
                 if res.rowcount == 0:
                     break
+            else:
+                break
```
rfk: Huh, was this just straight-up incorrect before, and would always execute the loop five times even when it succeeded the first time?

tokenserver/assignment/sqlnode/sql.py
```diff
@@ -475,6 +509,12 @@ def delete_user_record(self, service, uid):
         res = self._safe_execute(_DELETE_USER_RECORD, **params)
         res.close()
 
+    def migrate_user(self, service, uid, state):
+        """Set the migration state for a uid."""
+        params = {'service': service, 'uid': uid, 'state': state}
+        res = self._safe_execute(_MIGRATE_USER, **params)
+        res.close()
```
rfk: I don't really understand the intention behind this "migration state" field. Do you have an example of how you might use it in practice?

tokenserver/assignment/sqlnode/sql.py
```diff
@@ -384,6 +417,7 @@ def update_user(self, service, user, generation=None, client_state=None,
                 'nodeid': nodeid, 'generation': generation,
                 'keys_changed_at': keys_changed_at,
                 'client_state': client_state, 'timestamp': now,
+                'migration_state': ''
```
rfk: I don't really understand what "migration_state" is for, but I found myself expecting this to use the passed-in value rather than resetting to the empty string.

tokenserver/assignment/sqlnode/sql.py
```diff
@@ -315,9 +337,18 @@ def allocate_user(self, service, email, generation=0, client_state='',
...
+            'migration_state': ''
         }
         res = self._safe_execute(_CREATE_USER_RECORD, **params)
+        uid = res.lastrowid
+        # Update info if the user is selected to go to the spanner cluster:
+        if spanner_candidate and self.lucky_user(uid):
+            service = "spanner"
+            nodeid = params['nodeid'] = self._spanner_node_id
+            params['migration_state'] = "ORIGINAL"
+            node = self._spanner_entry
```
rfk: This overrides any explicit `node` argument provided by the caller, which seems like it could be confusing in future.

tokenserver/assignment/__init__.py
```diff
@@ -16,6 +21,8 @@ def get_user(self, service, email):
           * generation:  the last-seen generation number for that email
           * client_state:  the last-seen client state string for that email
           * old_client_states:  any previously--seen client state strings
+          * migration_state: one of None, "MIGRATING", "MIGRATED", where
+            a state of "MIGRATING" should cause a 501 return.
```
rfk: The way you use this field does not seem to match the way its described here.

tokenserver/tests/test_memorynode.ini
```diff
@@ -13,6 +13,9 @@ node = https://example.com
 token_duration = 3600
 node_type_patterns =
   example:*example*
+spanner_entry  = https://spanner.example.com
+spanner_node_id = 800
```
rfk: Something about naming these config entries after an implementation detail of the new backend seems unwise to me, but that's probably just my own unhealthy obsession with finding The Best Name for things.

tokenserver/tests/test_memorynode.py
```diff
@@ -45,3 +45,15 @@ def test_assignation(self):
...
+        settings = self.config.get_settings()
+        settings['tokenserver.migrate_new_user_percentage'] = 1
+        self.config.add_settings(settings)
+
+        user0 = self.backend.allocate_user(
+            DEFAULT_SERVICE, 'user0@example.com')
+        user1 = self.backend.allocate_user(
+            DEFAULT_SERVICE, 'user1@example.com')
+        self.assertEquals(user0['node'], settings['tokenserver.spanner_entry'])
+        self.assertEquals(user1['node'], settings['tokenserver.service_entry'])
```
rfk: This test feels like it could live in the more general `test_service.py` suite, rather than just being specific to the memory backend (which is the least important backend to test, since it's not the one we use in production).

I think it would also be worth adding a test to ensure the intended "stickiness" of this assignment, such as the user remaining assigned to spanner across a password change and password reset. Examples of similar tests can be found in `test_service.py`.

tokenserver/assignment/sqlnode/sql.py
jrconlin: As far as I can tell, yes. This always executed 5 times.

rfk: Nice catch :-)

jrconlin:
Thanks Ryan,

Part of this is probably due to my not quite understanding all the interaction points, so I absolutely expect some confusion.

The "migration_state" field serves several roles in my mind. One, it indicates a record that has been sent over to the Spanner system. Currently, this would only be new users, and the code only calls this in the case where a user is first assigned, but as older users are eventually migrated over, there would need to be some state to indicate that this user is no longer stored in MySQL (I am guessing this is the historical state you're concerned about). 

This is partly due to the fact that the migration work I was doing shows that transferring over users may result in less projected interruption time than locking a node, transferring the users, then routing them to the new service. Since the user identification information would remain the same in both mysql and spanner, and this data would be reflected in the data set here, it would be straightforward to honor any data deletion requests. 

In any case, that transition state needs to live somewhere, which means either a table or an additional column. Since the column doesn't require an index, does not impact existing data, and is tied to the same index that is used for the `users` table, adding the column made the most sense to me at the time. Very happy to move that state info somewhere else if it makes more sense.  

I chose to select users to send based on their ID partly to ease testing and to give some level of predictability about which users would be routed to the spanner system. I also agree that it's probably needlessly complicated since the `uid` for mysql is the row sequence number, which is not available until after the insertion, thus resulting in the immediate `MIGRATE_USER` call. That same call could be used later as part of the future user migration script.  Still, it's probably causing more problems than it's fixing and I'll simply opt for a random generator.

tokenserver/tests/test_memorynode.ini
jrconlin: Hey, I named the spanner selection function `lucky_user` so, yeah, I'm terrible at naming things. Very happy to accept better suggestions.

tokenserver/assignment/__init__.py
jrconlin: Yeah, I was thinking further ahead, and missed the state for a new user that is being sent to spanner as part of the migration test. I'll fix this to include the transition.

tokenserver/assignment/sqlnode/sql.py
jrconlin: Hmm, good point. Part of this is that I'm still not 100% sure how the tokenserver arguments are actually provided, and how often a new, unallocated user might be assigned to an existing node. I'll add logic to only possibly provision those users to spanner if no node is set.

tokenserver/assignment/sqlnode/sql.py
jrconlin: Ah, yeah, that's a bug. Thanks!

rfk:
> there would need to be some state to indicate that this user is no longer stored in MySQL

Sure, but the `nodeid` field indicates this already, but virtue of pointing to the spanner node :-)

> I chose to select users to send based on their ID partly to ease testing and to give some level of
> predictability about which users would be routed to the spanner system

Another option could be to use some deterministic function of the `email` parameter, e.g. `int(md5(email).hexdigest()[:8], 16)` or similar.

> This is partly due to the fact that the migration work I was doing shows that transferring over users
> may result in less projected interruption time than locking a node, transferring the users, then
> routing them to the new service. 
> Since the user identification information would remain the same in both mysql and spanner,

Ah, interesting - so is it still a possibility that we'll try to move user data over to spanner on the backend rather than requiring existing users to re-upload their data on the client?

I see the attraction of trying to keep `uid` stable across such a migration, but I think it's likely to cause unexpected side-effects elsewhere. I'd recommend creating a new record in the `users` table at the time of migration.

> and this data would be reflected in the data set here, it would be straightforward to honor
> any data deletion requests.

To honor data deletion requests, we'd have to remember which MySQL node they were assigned to prior to the migration, but the current version of `_MIGRATE_USER` overwrites that with the nodeid of the spanner node.

I'm not opposed to adding a `migration_state` column, just be be clear. I think it would work OK to create a new record for the user at the time of migration (marking their previous record as replaced) and set `migration_state` only on their current record. Then `update_user` could propagate that new column forward across password resets etc.

That said, I reckon we could safely defer adding it until we need to use it on existing users.

tokenserver/assignment/sqlnode/sql.py
jrconlin: Removed update_user

tokenserver/tests/test_service.py
```diff
@@ -806,8 +831,14 @@ def setUp(self):
         self.backend._safe_execute('delete from users')
         # Ensure the necessary service exists in the db.
         self.backend.add_service('sync-1.1', '{node}/1.1/{uid}')
+        self.backend.add_service('sync-1.5', '{node}/1.5/{uid}')
+        self.backend.add_service(
+            'spanner', 'https://spanner.example.com/1.5/{uid}')
```
rfk: Hrm. I'm surprised to see this working when using "spanner" as the service name here. Since the spanner node is "just another sync storage node" I would expect these tests to all use `"sync-1.5"` as the service name, and for you to be adding the spanner node as `self.backend.add_node("sync-1.5", "https://spanner.example.com/1.5/{uid}")`.

rfk: The "service name" parameter is really weird in this codebase, because it's a legacy hangover from a time when we thought we'd have multiple different services relying on Tokenserver for authentication. In practice we've only ever had one service, "sync-1.5", and we should expect to only ever have that one service in the future. If it wasn't such a wide-ranging change I'd suggest we go through and remove the `service` argument to all calls into the node-assignment backend, since it's totally useless to us in practice.

tokenserver/assignment/sqlnode/sql.py
```diff
@@ -596,10 +608,16 @@ def unassign_node(self, service, node, timestamp=None, nodeid=None):
...
 
-    def get_best_node(self, service):
+    def get_best_node(self, service, email):
         """Returns the 'least loaded' node currently available, increments the
         active count on that node, and decrements the slots currently available
         """
+        if self.lucky_user(email):
+            return (
+                self.settings.get("spanner_node_id", 800),
+                self.settings.get("spanner_entry", None)
```
rfk: Reading both the node_id and its corresponding URL from settings seems redundant, and it might mask configuration errors (such as if we accidentally set `spanner_node_id` to a value that exists in the database as a MySQL node).

What do you think about requiring ops to add an explicit entry in the `nodes` table to represent the spanner node? (There is a `tokenserver/scripts/add_node.py` script that can be used for this purpose, for reference). This would ensure it cannot conflict with any MySQL nodes. We could then have *just* the spanner node URL in the config file, and look up the appropriate id at runtime by querying the `nodes` table.  

This would also help ensure that we respect the existing semantics of the `service` parameter, as mentioned in my comment on the tests. It's useless, but I think there's value in maintaining consistent behaviour.

rfk: Also, defaulting to `800` here seems risky, unless something prevents creating a MySQL node in the db with id `800`.

tokenserver/assignment/memorynode.py
```diff
@@ -19,6 +20,7 @@ def __init__(self, service_entry=None, **kw):
         self._service_entry = service_entry
         self._users = {}
         self._next_uid = 1
+        self.settings = get_current_registry().settings
```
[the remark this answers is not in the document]
rfk: Actually, it's not clear to me why this needs to access the settings differently than the SQL backend. I think it should be able to obtain settings via `**kw` argument in the same way that the SQL backend does, since the backend should be loaded via abstract interface [here](https://github.com/mozilla-services/tokenserver/blob/daf36fb991e0b3fe34dd21d62c20d192f968ff44/tokenserver/__init__.py#L40)

Perhaps something in the tests is initializing the `MemorynodeAssignmentBackend` differently? If so, I think we should find and fix it there rather than handle settings differently between the two backends.

tokenserver/tests/test_service.py
```diff
@@ -791,6 +791,31 @@ def test_node_type_is_returned_in_response(self):
         res = self.app.get('/1.0/sync/1.1', headers=headers, status=200)
         self.assertEqual(res.json['node_type'], 'example')
 
+    def test_migrate_new_user(self):
+        EMAIL0 = "abO-test@example.com"
+        EMAIL1 = "abT-test@example.com"
+        settings = self.backend.settings
+        # the two platfroms differ in how the settings are labeled.
```
rfk: Per my comment on the memory-node backend, I don't think this should be necessary. If you try to make the memory backend access the settings through `**kw` in the same way as the SQL backend, what fails?

tokenserver/assignment/sqlnode/sql.py
rfk: > We could then have just the spanner node URL in the config file, and look up the appropriate id at runtime by querying the nodes table.

Actually I'm not sure we have an appropriate index for this lookup. It may be fine to do a table-scan lookup once at load time, or maybe we could put *just* the id in the config file after inserting it into the db and finding out what it is.

tokenserver/assignment/memorynode.py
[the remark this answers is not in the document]
jrconlin: hrm. I think part of the problem is that `get_current_registry()` is a Pyramid function that returns the settings defined within the Pyramid framework. Each setting is namespaced. 

The `kw` argument, however, is a local calling parameter, so the settings aren't namespaced. 

This is a bit more complicated because the nose tests don't always specify either the kw or the pyramid registry. I'll try to normalize them, but might be a bit more than the scope of this PR.

tokenserver/assignment/__init__.py
```diff
@@ -4,6 +4,11 @@
 class INodeAssignment(Interface):
     """Interface definition for backend node-assignment db."""
 
+    def lucky_user(self):
```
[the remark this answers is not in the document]
jrconlin: better than what I came up with. I'll trim this down to `allocate_to_spanner` since that parses a bit better.

tokenserver/assignment/sqlnode/sql.py
```diff
@@ -304,18 +306,27 @@ def get_user(self, service, email):
         finally:
             res.close()
 
+    def lucky_user(self, email):
+        pick = ord(hashlib.sha1(email.encode()).digest()[0])
```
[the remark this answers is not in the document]
jrconlin: Heh, it's why I didn't do `md5` like you suggested.

/me _thinks about the sha1() collisions article in PoC:GTFO..._

tokenserver/assignment/sqlnode/sql.py
```diff
@@ -596,10 +608,16 @@ def unassign_node(self, service, node, timestamp=None, nodeid=None):
...
+    def get_best_node(self, service, email):
         """Returns the 'least loaded' node currently available, increments the
         active count on that node, and decrements the slots currently available
         """
+        if self.lucky_user(email):
+            return (
+                self.settings.get("spanner_node_id", 800),
+                self.settings.get("spanner_entry", None)
+            )
+
```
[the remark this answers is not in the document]
jrconlin: Huh, i'd figure that those users would be basically indistinguishable from new users in that we'd toss their previous state, but yeah, the would become candidates for the new system as well.

tokenserver/tests/test_service.py
jrconlin: Ah, to be honest, I wasn't quite sure what service actually equated to. I could see there being a new "spanner" service which doesn't have a node per se, since, in theory, spanner can handle an infinite amount of entries. I'm actually wondering how this might play out for ops. 

I could see a far future where we have a dedicated 'spanner' service or have clients `opt-in` themselves, but, I agree, this might be more effort than is wise.

tokenserver/assignment/sqlnode/sql.py
jrconlin: As I understand it from ops, `spanner` is already set as a node_id `800`, so we're ok with hardcoding it for now. I'll reconfirm with @Micheletto when he gets back from PTO.

I'm 100% OK with ops controlling the node table (I'm half tempted to suggest they also control the services table too based on the above comment)

tokenserver/tests/test_service.py
jrconlin: It's not a full on fail, just a mis-match between the various ways that these values are set and read. I've added a `settings` property to try and normalize the values that are being passed in regardless of how. For what it's worth, it also gave me a cleaner way to do test settings.

tokenserver/assignment/sqlnode/sql.py
rfk: Ok, great! I reckon we should be able to put *just* the spanner node id in the config file then, and use the existing logic to reach the endpoint URL out of the database from the `nodes` table.

rfk: Re-reading this, I see that you probably need the `spanner_entry` config option in order to support the memory backend. It still seems like a footgun to support it in the sql backend. (And TBH I don't recall why we need a separate memory backend at all versus figuring out how to use a sqlite `:memory:` database).

tokenserver/assignment/sqlnode/sql.py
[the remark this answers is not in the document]
rfk: We keep their old state around after a password reset, both so that it can be dealt with by the various background cleanup jobs, and so that we can prevent certain kinds of client misbehavior (e.g. the "you tried to use an old client_state value" errors). But I think it's fine to put them in the same bucket as new users here.

tokenserver/assignment/__init__.py
```diff
@@ -4,6 +4,11 @@
 class INodeAssignment(Interface):
     """Interface definition for backend node-assignment db."""
 
+    def allocate_to_spanner(self, email):
```
rfk: extremely picky nit: this method implements a predicate but is named like a verb, and behaves very differently to the similarly-named `allocate_user` method. Would you consider something like `should_allocate_to_spanner` instead?

I also don't think it needs to be part of the public API definition, but don't feel strongly if you want to keep it here.

tokenserver/assignment/sqlnode/sql.py
```diff
@@ -233,6 +237,22 @@ def __init__(self, sqluri, create_tables=False, pool_size=100,
...
+        # `tokenserver.` settings from the pyramid settings
+        # registry, remove the namespace prefix so that they
+        # match the values that should be passed in as *kw
+        # to the __init__()
+        settings = dict(
+           map(lambda (k, v): (
+               k.replace('tokenserver.', ''), v),
+               filter(lambda e: e[0].startswith('tokenserver.'),
+                      (get_current_registry().settings or {}).items()))
+        ) or self._settings
```
rfk: Should this `or` be the other way around, i.e. `self._settings or dict(...)`?

I'll also note that this is going to iterate all the keys of `get_current_registry().settings` multiple times per request, once for every access to `self.settings`. Probably fine in practice but it seems wasteful.

tokenserver/tests/test_service.py
rfk: > It's not a full on fail, just a mis-match between the various ways that these values are set
> and read.

I guess I don't understand why these values are getting set in more than one way; every creation of an `INodeAssignment` backend should be being routed through the call to `load_and_register("tokenserver", config)` in `tokenserver/__init__.py`, but maybe that's not happening in all cases.

rfk:
Is it important that we *prevent* existing users from being organically node-assigned to the spanner node?

tokenserver/tests/test_service.py
jrconlin: Ah, apparently, with the unified test suite we run, the MySQL side will fail if a "spanner" service is not defined, even though it may never call "spanner". Rather than add additional complexity to the test or functional code, I'll just add a comment to the `setUp()` function justifying why it's present.

tokenserver/assignment/sqlnode/sql.py
jrconlin: Well, this would favor the thread local registry settings over the kw values passed in during the Node init. If I remember, you said that the registry settings could change at any point, even after init. I agree that it's a bit wasteful. If these aren't as dynamic as I thought, I'd prefer to do this sort of stuff once and then cache the result.

tokenserver/tests/test_service.py
jrconlin: Also, removed this comment, since the setting normalize function invalidates it.

jrconlin: Ok, having dug into this a bit more:
The problem is that there are two different pools from which "settings" can be derived, and they overlap a bit. During the init, `kw` values are extracted from the pyramid settings and the various plugins are called with those `kw` values, trimmed of their prefixes. Then pyramid creates some thread local values that are the original settings values with prefixes. 

The tokenserver code [sometimes calls the thread local settings](https://github.com/mozilla-services/tokenserver/blob/daf36fb991e0b3fe34dd21d62c20d192f968ff44/tokenserver/assignment/memorynode.py#L28), which as I understand, could change dynamically. Mind you, this sort of call doesn't happen regularly, and the one more problematic one I highlighted earlier probably could be converted to a `kw` init to be consistent with other forms. (There's another time the code [calls to the thread local settings](https://github.com/mozilla-services/tokenserver/blob/2ade4a7c6ab8658a11a7674c47ea74afdc7ad2bb/tokenserver/verifiers.py#L32) registry, but that one appears to be unrelated to configuration.

If the assignment node configurations aren't as dynamic, I can simplify this code a good deal, remove the thread local transmogrification bits, etc.

tublitzed:
> Is it important that we prevent existing users from being organically node-assigned to the spanner node?

@rfk - We've been planning the rollout in two stages; [first new users only](https://docs.google.com/document/d/1ULkyVWBeIt0zYo4AmF5aUI1kRDKCPQQaXVZv1w1oj_M/edit#) (less risky, no data to lose), and then [migrating existing users](https://docs.google.com/document/d/1nXkq5KTWuV86pYMwaZ_zZxDZ4JNyIMdy-pYzvxIsPO0/edit#heading=h.ly6hgbd1hc13). So, ideally yes; we'd like to be able to route only new users first. I don't think it would be catastrophic if a small number of existing users were routed over, but since that's the riskiest bit here it would be good to be able to ensure we've migrated their data first to reduce the risk of them connecting on some new device that doesn't have their sync data and losing access.

tokenserver/assignment/sqlnode/sql.py
rfk: > you said that the registry settings could change at any point

I do recall saying this, and it's true, but as mentioned in slack I think I've convinced myself that our use of `get_current_registry()` is entirely unnecessary in practice and could be safely removed.

(There's a whole bunch of indirection that happens via component registration in pyramid, which results in the app always loading the `INodeAssignment` backend from the current registry, and the backend from the current registry should always have been created using the settings from that registry, so there's no reason for us to check them on every request).

tokenserver/tests/test_service.py
rfk: > the MySQL side will fail if a "spanner" service is not defined, even though it may never call "spanner".

I can't imagine what could be causing this in practice, but I can't argue with "makes the tests fail" :-)

jrconlin: Moving the test to sql only seems to solve this, so happily dropping the nonce calls.

tokenserver/assignment/memorynode.py
```diff
@@ -19,14 +19,14 @@ def __init__(self, service_entry=None, **kw):
...
+        self.settings = kw or {}
 
     @property
     def service_entry(self):
         """Implement this as a property to have the context when looking for
         the value of the setting"""
         if self._service_entry is None:
-            settings = get_current_registry().settings
-            self._service_entry = settings.get('tokenserver.service_entry')
+            self._service_entry = self.settings.get('service_entry')
```
rfk: I don't think this will ever do anything in practice - since the constructor takes `service_entry` as a positional argument, it should never appear in `**kw` and hence never appear in `self.settings`. I think we can just remove this property and use `self._service_entry` directly.

tokenserver/assignment/memorynode.py
```diff
@@ -39,6 +39,14 @@ def get_user(self, service, email):
         except KeyError:
             return None
 
+    def should_allocate_to_spanner(self, email):
```
rfk: This is no longer used in the memory backend, and can be removed. (Alternately, if you want to keep it in the public interface, we could consider moving it into some sort of shared helper. I don't feel particularly strongly about it, just don't like seeing unused code).

tokenserver/assignment/sqlnode/sql.py
```diff
@@ -224,6 +234,7 @@ def __init__(self, sqluri, create_tables=False, pool_size=100,
         self.services = get_cls('services', _Base)
         self.nodes = get_cls('nodes', _Base)
         self.users = get_cls('users', _Base)
+        self.settings = kw or {}
```
rfk: `self.settings` seems unused in this file.

tokenserver/tests/assignment/test_sqlnode.py
```diff
@@ -29,6 +29,8 @@ def setUp(self):
         self.backend.add_service('sync-1.5', '{node}/1.5/{uid}')
         self.backend.add_service('queuey-1.0', '{node}/{service}/{uid}')
         self.backend.add_node('sync-1.0', 'https://phx12', 100)
+        self.backend.add_node('sync-1.5', 'https://phx11', 100)
+        self.backend.migrate_new_user_percentage = 0
```
rfk: If I delete these two lines, all of the tests still seem to pass. Are they necessary, perhaps only under some test configurations?

tokenserver/tests/test_service.py
```diff
@@ -806,8 +809,14 @@ def setUp(self):
         self.backend._safe_execute('delete from users')
         # Ensure the necessary service exists in the db.
         self.backend.add_service('sync-1.1', '{node}/1.1/{uid}')
+        self.backend.add_service('sync-1.5', '{node}/1.5/{uid}')
         # Ensure we have a node with enough capacity to run the tests.
-        self.backend.add_node('sync-1.1', 'https://example.com', 100)
+        self.backend.add_node('sync-1.1', self.mysql_node, 100)
+        self.backend.add_node('sync-1.5', self.mysql_node, 100)
+        # Ensure we have a spanner node, but give it not capacity
```
rfk: type (which I'm sure was my own): "not capacity" -> "no capacity".

tokenserver/tests/test_service.py
```diff
@@ -816,6 +825,18 @@ def tearDown(self):
         self.backend._safe_execute('delete from users')
         super(TestServiceWithSQLBackend, self).tearDown()
 
+    def test_assign_new_users_to_spanner(self):
+        self.backend.migrate_new_user_percentage = 1
+        # These emails are carefully selected os that the first is assigned
```
rfk: typo: "os" -> "so"

tokenserver/tests/test_sql.ini
```diff
@@ -15,6 +15,9 @@ secrets.master_secrets = "abcdef"
                          "123456"
 node_type_patterns =
   example:*example.com
+spanner_entry  = https://spanner.example.com
```
rfk: This `spanner_entry` key is no longer used.

tokenserver/tests/test_memorynode.ini
```diff
@@ -13,6 +13,9 @@ node = https://example.com
...
+migrate_new_user_percentage=0
```
rfk: These config entries are no longer used by the memory backend.

tokenserver/assignment/sqlnode/sql.py
```diff
@@ -655,15 +691,17 @@ def get_best_node(self, service):
...
-        con = self._safe_execute(query, close=True)
-        con.close()
+        if not send_to_spanner:
+            # Update the node to reflect the new assignment.
+            # This is a little racy with concurrent assignments, but no big
+            # deal.
+            where = [nodes.c.service == service, nodes.c.node == node]
+            where = and_(*where)
+            fields = {'available': self._sqlfunc_max(nodes.c.available - 1, 0),
+                      'current_load': nodes.c.current_load + 1}
```
rfk: I can see a case for not updating `available` on the spanner node, but tracking `current_load` might still be useful to us in the future. But I don't feel strongly either way.

jrconlin: Yeah, I can kinda see that. adding it back in.

tokenserver/tests/assignment/test_sqlnode.py
jrconlin: probably in there for completion sake, but yeah, not really needed.

tokenserver/assignment/sqlnode/sql.py
jrconlin: Ah, good point.

tokenserver/assignment/memorynode.py
jrconlin: nah, not needed.

tokenserver/assignment/memorynode.py
jrconlin: true.

PR #164, opened 2019-12-30
Some suggestions for simplifying config handling in #163

rfk:
@jrconlin I was puzzled by https://github.com/mozilla-services/tokenserver/pull/163#discussion_r361885306 so decided to dig in a bit and see what was going on. In the process I came to the opinion that you shouldn't need to add spanner handling to the `MemoryNodeAssignmentBackend` at all - the whole premise of that backend is that it's a very simple setup where everyone is on a single node, and adding the complexity of a new second node doesn't seem to provide us any value. I don't think anyone would ever actually use it.

So...here's a proposal to simplify the config handling in #163, by removing spanner from the memory backend and configuring it exclusively via `**kw` in the sql backend. I also rejiggered some tests around to match.

Please feel free to take it or leave it, but I wanted to get my head around what was happening with all the config, and the easiest way to do that was with code.

### DIFF
--- a/tokenserver/assignment/memorynode.py
+++ b/tokenserver/assignment/memorynode.py
@@ -1,6 +1,3 @@
-import hashlib
-
-from pyramid.threadlocal import get_current_registry
 from zope.interface import implements
 
 from tokenserver.assignment import INodeAssignment
@@ -20,32 +17,6 @@ class MemoryNodeAssignmentBackend(object):
         self._service_entry = service_entry
         self._users = {}
         self._next_uid = 1
-        self._test_settings = {}  # unit test specific overrides
-        self._settings = kw or {}
-
-    @property
-    def settings(self):
-        # Normalize the various settings by picking out the
-        # `tokenserver.` settings from the pyramid settings
-        # registry, remove the namespace prefix so that they
-        # match the values that should be passed in as *kw
-        # to the __init__()
-        settings = dict(
-           map(lambda (k, v): (
-               k.replace('tokenserver.', ''), v),
-               filter(lambda e: e[0].startswith('tokenserver.'),
-                      (get_current_registry().settings or {}).items()))
-        ) or self._settings
-        settings.update(self._test_settings)
-        return settings
-
-    @property
-    def service_entry(self):
-        """Implement this as a property to have the context when looking for
-        the value of the setting"""
-        if self._service_entry is None:
-            self._service_entry = self.settings.get('service_entry')
-        return self._service_entry
 
     def clear(self):
         self._users.clear()
@@ -57,28 +28,16 @@ class MemoryNodeAssignmentBackend(object):
         except KeyError:
             return None
 
-    def should_allocate_to_spanner(self, email):
-        """use a simple, reproducable hashing mechanism to determine if
-        a user should be provisioned to spanner. Does not need to be
-        secure, just a selectable percentage."""
-        return ord(hashlib.sha1(email.encode()).digest()[0]) < (
-            256 * (self.settings.get(
-                    'migrate_new_user_percentage', 0) * .01))
-
     def allocate_user(self, service, email, generation=0, client_state='',
                       keys_changed_at=0, node=None):
         if (service, email) in self._users:
             raise BackendError('user already exists: ' + email)
         if node is not None and node != self.service_entry:
             raise ValueError("unknown node: %s" % (node,))
-        if self.should_allocate_to_spanner(email):
-            service_entry = self.settings.get('spanner_entry')
-        else:
-            service_entry = self.service_entry
         user = {
             'email': email,
             'uid': self._next_uid,
-            'node': service_entry,
+            'node': self._service_entry,
             'generation': generation,
             'keys_changed_at': keys_changed_at,
             'client_state': client_state,
--- a/tokenserver/assignment/sqlnode/sql.py
+++ b/tokenserver/assignment/sqlnode/sql.py
@@ -12,7 +12,6 @@ import math
 import traceback
 import hashlib
 from mozsvc.exceptions import BackendError
-from pyramid.threadlocal import get_current_registry
 
 from sqlalchemy.sql import select, update, and_
 from sqlalchemy.ext.declarative import declarative_base
@@ -190,12 +189,12 @@ class SQLNodeAssignment(object):
     def __init__(self, sqluri, create_tables=False, pool_size=100,
                  pool_recycle=60, pool_timeout=30, max_overflow=10,
                  pool_reset_on_return='rollback', capacity_release_rate=0.1,
+                 spanner_node_id=None, migrate_new_user_percentage=0,
                  **kw):
         self._cached_service_ids = {}
         self.sqluri = sqluri
         if pool_reset_on_return.lower() in ('', 'none'):
             pool_reset_on_return = None
-        self._test_settings = {}  # unit test specific overrides
 
         # Use production-ready pool settings for the MySQL backend.
         # We also need to work around mysql using "LEAST(a,b)" and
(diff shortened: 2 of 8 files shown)


tokenserver/assignment/memorynode.py
```diff
@@ -20,32 +17,6 @@ def __init__(self, service_entry=None, **kw):
...
-               filter(lambda e: e[0].startswith('tokenserver.'),
-                      (get_current_registry().settings or {}).items()))
-        ) or self._settings
-        settings.update(self._test_settings)
-        return settings
-
-    @property
-    def service_entry(self):
-        """Implement this as a property to have the context when looking for
-        the value of the setting"""
```
rfk: I have no idea why we ever needed to fall back to `get_current_registry` for this property in the memory backend, but removing it didn't cause any tests to fail, so I figure it's some historical legacy that's no longer required.

tokenserver/assignment/sqlnode/sql.py
```diff
@@ -634,35 +625,34 @@ def get_best_node(self, service, email):
...
-        query = select([nodes]).where(and_(*where))
-
-        if self._is_sqlite:
-            # sqlite doesn't have the 'log' funtion, and requires
-            # coercion to a float for the sorting to work.
-            query = query.order_by(nodes.c.current_load * 1.0 /
-                                   nodes.c.capacity)
+        if self.allocate_to_spanner(email):
+            # Some users get allocated directly to the new spanner node.
+            query = query.where(nodes.c.id == self._spanner_node_id)
```
rfk: This looks up the metadata for the spanner node in the db, rather than having to duplicate it all into the config file. We were already hitting the db here in a pre-spanner world to find the best node, so I figure there won't be any perf impact to doing this lookup on every request. We could optimize to only do it once if desired.

tokenserver/tests/assignment/test_sqlnode.py
```diff
@@ -28,13 +28,8 @@ def setUp(self):
         self.backend.add_service('sync-1.0', '{node}/1.0/{uid}')
         self.backend.add_service('sync-1.5', '{node}/1.5/{uid}')
         self.backend.add_service('queuey-1.0', '{node}/{service}/{uid}')
-        self.backend.add_service('spanner',
-                                 'https://spanner.example.com/1.5/{uid}')
         self.backend.add_node('sync-1.0', 'https://phx12', 100)
         self.backend.add_node('sync-1.5', 'https://phx11', 100)
-        self.backend.add_node('spanner',
-                              'https://spanner.example.com/1.5/{uid}', 9000)
-        self.backend._migrate_new_user_percentage = 0
```
rfk: I don't think these were used for anything. We could consider adding additional tests in this file, e.g. to confirm that percentage-based selection works as we expect when you change the percentage.

tokenserver/tests/test_service.py
```diff
@@ -825,13 +807,29 @@ def setUp(self):
...
+        EMAIL1 = "abT-test@example.com"
+
+        user0 = self.backend.allocate_user("sync-1.5", EMAIL0)
+        user1 = self.backend.allocate_user("sync-1.5", EMAIL1)
+        self.assertEquals(
+            user0['node'],
+            'https://spanner.example.com')
+        self.assertEqual(
+            user1['node'],
+            'https://example.com')
```
rfk: This test is now specific to the sql backend.

tokenserver/tests/test_sql.ini
```diff
@@ -15,9 +15,8 @@ secrets.master_secrets = "abcdef"
                          "123456"
 node_type_patterns =
   example:*example.com
-spanner_entry  = https://spanner.example.com
 spanner_node_id = 800
-migrate_new_user_percentage=0
+migrate_new_user_percentage=1
```
rfk: By setting this in the config file rather than in code, we test that the loading-from-config-file logic also works as expected.

tokenserver/assignment/sqlnode/sql.py
```diff
@@ -701,7 +693,7 @@ def get_best_node(self, service, email):
         # This is a little racy with concurrent assignments, but no big deal.
         where = [nodes.c.service == service, nodes.c.node == node]
         where = and_(*where)
-        fields = {'available': nodes.c.available - 1,
+        fields = {'available': self._sqlfunc_max(nodes.c.available - 1, 0),
```
rfk: This change has the advantage of correctly maintaining `current_load` for the spanner node, but the risk of underflowing `available` because we ignore its `capacity`. Don't allow it to underflow.

tokenserver/assignment/sqlnode/sql.py
```diff
@@ -554,13 +544,16 @@ def add_node(self, service, node, capacity, **kwds):
...
-            insert into nodes (service, node, available, capacity,
+            insert into nodes (id, service, node, available, capacity,
                             current_load, downed, backoff)
-            values (:service, :node, :available, :capacity,
+            values (:nodeid, :service, :node, :available, :capacity,
                     :current_load, :downed, :backoff)
             """),
-            service=service, node=node,
-            capacity=capacity, available=available,
+            nodeid=kwds.get('nodeid', None),
```
rfk: This is to support the tests, by inserting the spanner node with a known `nodeid`.

tokenserver/tests/test_sql.ini
jrconlin: hrm, this will fire on every sql test, though. Going to move this to the actual test to limit future swearing.

tokenserver/assignment/sqlnode/sql.py
jrconlin: well, since this is updating the node info, and spanner is a special case, treating spanner as a special case and not updating the nodes when we send an account.

tokenserver/assignment/sqlnode/sql.py
jrconlin: future nit: We should be consistent about `kw` `kwds` etc.

tokenserver/assignment/sqlnode/sql.py
```diff
@@ -224,10 +225,16 @@ def __init__(self, sqluri, create_tables=False, pool_size=100,
         else:
             from tokenserver.assignment.sqlnode.schemas import get_cls
 
+        self._spanner_node_id = spanner_node_id
+        self._migrate_new_user_percentage = migrate_new_user_percentage
+        if self._migrate_new_user_percentage > 0:
+            if self._spanner_node_id is None:
+                raise ValueError("Must set spanner_node_id when migrating "
+                                 "users to spanner")
```
jrconlin: @pjenvey would not approve. ;)

tokenserver/tests/test_sql.ini
rfk: Does it matter though? None of the behaviour of the other tests should be affected by whether the user goes to the (fake) spanner node or not.

Alternately, maybe we should move this into the backend tests under `tests/assignment/` if it's specific to a particular backend (which may be where you had it originally).

tokenserver/assignment/sqlnode/sql.py
rfk: Well now I'm curious as to why :-)

tokenserver/tests/test_sql.ini
jrconlin: Hi Ryan! Have you met my good friend Murphy? He likes to drop by late at night in the future while trying to figure out why some tests are going weird. ;)

Murphy and I go way, way back and I've learned to look for unlocked doors, loose window jams, and unsecured floor boards he might use to get in.

tokenserver/assignment/sqlnode/sql.py
jrconlin: He prefers that exceptions like this be package specific and not use existing exception types when possible. (So, something like `ConfigException` rather than `ValueError`) I've kinda learned to agree since using a custom allows you a good deal more flexibility in the future with things like error reporting and potential mitigation.

rfk: That's fair; come to think of it I usually prefer to use `RuntimeError` or `AssertionError` for things like this, since they're less likely to be accidentally caught and ignored.

rfk:
Closing since JR incorporated bits of this into the upstream PR.

PR #166, opened 2020-01-06
feat: add dynamic_settings table for on-the-fly ops config

### LINKED ISSUE #165: Provide a "dynamic_settings" table in SQL for ops
Ops has noted that updating a setting in a SQL table does not require a redepoyment. For certain settings, it may be more useful to have a "settings" table that they can modify on the fly.

jrconlin:
Allows ops to modify `dynamic_settings` `migrate_new_user_percentage`
without requiring a redeployment.

Closes #165

### DIFF
--- a/tokenserver/assignment/sqlnode/sql.py
+++ b/tokenserver/assignment/sqlnode/sql.py
@@ -192,6 +192,8 @@ where
     setting = :setting
 """)
 
+MIGRATION_CACHE_LIFESPAN = 300
+
 
 class SQLNodeAssignment(object):
 
@@ -203,7 +205,7 @@ class SQLNodeAssignment(object):
                  spanner_node_id=None, migrate_new_user_percentage=0,
                  **kw):
         self._cached_service_ids = {}
-        self._cache_migration_ttl = 0
+        self._migration_percentage_cache_ttl = 0
         self.sqluri = sqluri
         if pool_reset_on_return.lower() in ('', 'none'):
             pool_reset_on_return = None
@@ -331,7 +333,7 @@ class SQLNodeAssignment(object):
     def get_migration_percent(self):
         """get a cached Ops controllable percentage value for the number of
         new users to migrate to Spanner."""
-        if self._cache_migration_ttl > time.time():
+        if self._migration_percentage_cache_ttl > time.time():
             return self.migrate_new_user_percentage
         default = self.migrate_new_user_percentage or 0
         try:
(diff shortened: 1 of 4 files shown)


tokenserver/assignment/sqlnode/sql.py
```diff
@@ -192,6 +203,7 @@ def __init__(self, sqluri, create_tables=False, pool_size=100,
                  spanner_node_id=None, migrate_new_user_percentage=0,
                  **kw):
         self._cached_service_ids = {}
+        self._cache_migration_ttl = 0
```
rfk: naming nit: as a reader, the name of this property doesn't help me understand how it's going to be used. Does it control a "cache migration" feature that I haven't encountered yet? Is it the number of seconds for which the migration should live? I suggest something like `_cached_migration_exp_at`, but I don't feel strongly enough about it to bikeshed in any detail 😀

tokenserver/assignment/sqlnode/sql.py
```diff
@@ -314,13 +328,35 @@ def get_user(self, service, email):
...
+                .format(ex, default)
+            )
+        return self.migrate_new_user_percentage
+
     def should_allocate_to_spanner(self, email):
         """use a simple, reproducable hashing mechanism to determine if
         a user should be provisioned to spanner. Does not need to be
         secure, just a selectable percentage."""
-        if self.migrate_new_user_percentage:
+        migrate = self.get_migration_percent()
```
rfk: When the timer on the cached value expires, we may end up with several concurrent requests trying to read and update the value at once. That's...probably fine i guess? I wonder if it's worth mentioning in a comment here for future readers.

jrconlin:
This was mostly requested by Ops. In order to change the percentage of new users we opt in to Spanner, they currently need to redeploy the entire package in order to change the one option. Reading the value from a database means that they do not need to redeploy.

rfk:
Let's make sure we at least set an ACL on this table in prod so that the app can't write to it.

tokenserver/assignment/sqlnode/sql.py
jrconlin: I don't think we should see any updates, but it is possible that a few threads have different values while the various TTLs live out. I think that's fine since we're talking about a reasonably short window. Things don't have to be exact, but it'll be faster than a system redeploy.

jrconlin:
This patch will check the `dynamic_settings` for a value marked as `migrate_new_user_percentage` which will have the same effect as setting `--migrate_new_user_percentage` as a CLI option or config file setting. 

As noted in [Ryan's Comment](https://github.com/mozilla-services/tokenserver/pull/166#issuecomment-574355437) The `dynamic_settings` table should only be write accessible by the authorized personnel in production.

PR #169, opened 2020-01-10
purge_old_records: add --max-offset option and some additional logging statements

jrgm:
r? - @rfk

I was working with the purge task in production to get it to run more effectively, and noticed a couple of things.

1. the default max-per-loop of 10 meant that each task was spending a large chunk of each cycle doing the select, and spinning a lot of database cpu. So I'll be using a max-per-loop of ~1000 going forward.

2. Due to the current select for old_records, each purge task winds up working on the same set of old_records. So this PR adds a --max-offset option to the task. The idea is that if each instance's task picks a random offset < max_offset, then it will be unlikely to have any 2 instances working on the same set of records. I'm thinking of running something like max-per-loop=1000 and max-offset=100000. The offset will add some additional cost to each select, so I'll have to play with those numbers to get the right tradeoff. But one thing I'm unclear about: the current query does `order by replaced_at desc, uid desc`. Is there some importance to that ordering? I.e., if I'm skipping to the middle of that list with an offset does that break some expectation?

This PR also adds a couple of logging statements I was interested in being able to see.

(Note: one gotcha with `--max-offset` is that if the number of purgeable rows is << max_offset, then **no** rows will be purged. I didn't try to defend against that in this PR. When the backlog is down to a manageable level I'll change max_offset to be zero (or maybe 10000 to keep a little bit of scatter). Or maybe not change it all: enough instances will get work to do by chance to keep to backlog under control).

### DIFF
--- a/tokenserver/scripts/purge_old_records.py
+++ b/tokenserver/scripts/purge_old_records.py
@@ -35,12 +35,17 @@ logger = logging.getLogger("tokenserver.scripts.purge_old_records")
 
 
 def purge_old_records(config_file, grace_period=-1, max_per_loop=10,
-                      request_timeout=60):
+                      max_offset=0, request_timeout=60):
     """Purge old records from the assignment backend in the given config file.
 
     This function iterates through each storage backend in the given config
     file and calls its purge_expired_items() method.  The result is a
     gradual pruning of expired items from each database.
+
+    `max_offset` is used to select a random offset into the list of purgeable
+    records. With multiple tasks running concurrently, this will provide each
+    a (likely) different set of records to work on. A cheap, imperfect
+    randomization.
     """
     logger.info("Purging old user records")
     logger.debug("Using config file %r", config_file)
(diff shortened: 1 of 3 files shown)


rfk:
> the current query does order by replaced_at desc, uid desc.
> Is there some importance that that ordering

I don't think it will affect the correctness of the script; most likely this is just trying to return the results in a consistent order to simplify testing.

tokenserver/scripts/purge_old_records.py
```diff
@@ -53,16 +53,21 @@ def purge_old_records(config_file, grace_period=-1, max_per_loop=10,
             logger.debug("Purging old user records for service: %s", service)
             # Process batches of <max_per_loop> items, until we run out.
             while True:
+                offset = random.randint(0, max_offset)
```
rfk: In 3 months time I'm not going to remember why we do this; could you please add a short paragraph to the docstring of this function, along the lines of what you explained in the PR here?

jrgm:
I added a note in the docstring, and an inline comment about <<< max_offset.

PR #171, opened 2020-01-22
get_old_user_records: restrict result set to exclude downed nodes

### LINKED ISSUE #170: How to further fix to purge_old_records task (?)
I was running the updated purge_old_records task in production, and the new changes were working fine, until it began hitting "rich veins" of downed nodes.

The task would pull up 1000 candidate rows and then immediately discard them all for downed nodes and re-select another result set. This would peg the unacceptably cpu on the database server.

I tried this hack to get around this problem:
```
diff --git a/tokenserver/scripts/purge_old_records.py b/tokenserver/scripts/purge_old_records.py
index dcb248f..78fcd35 100644
--- a/tokenserver/scripts/purge_old_records.py
+++ b/tokenserver/scripts/purge_old_records.py
@@ -79,6 +79,12 @@ def purge_old_records(config_file, grace_period=-1, max_per_loop=10,
                         delete_service_data(config, service, row,
                                             timeout=request_timeout)
                         backend.delete_user_record(service, row.uid)
+                    else:
+                        logger.info("Punting on this row: uid %s on %s (%d)",
+                                    row.uid, row.node, row.downed)
+                        # Essentially pause for 60 seconds if all the nodes
+                        # in these return rows are downed.
+                        time.sleep(60.0/len(rows))
                 if len(rows) < max_per_loop:
                     break
     except Exception:
```
... but that settled into a state where it never found any records to purge (but didn't peg the database cpu).

I tried increasing the limit and offset and occasionally it found pockets to purge, but really wasn't making a lot of progress.

I think this may need a change in the _GET_OLD_USER_RECORDS_FOR_SERVICE query. Something like:
```
diff --git a/tokenserver/assignment/sqlnode/sql.py b/tokenserver/assignment/sqlnode/sql.py
index b602f06..9c7166f 100644
--- a/tokenserver/assignment/sqlnode/sql.py
+++ b/tokenserver/assignment/sqlnode/sql.py
@@ -112,15 +112,19 @@ where
 
 _GET_OLD_USER_RECORDS_FOR_SERVICE = sqltext("""\
 select
     uid, email, generation, keys_changed_at, client_state,
     nodes.node, nodes.downed, created_at, replaced_at
 from
-    users left outer join nodes on users.nodeid = nodes.id
+    users
+use index (replaced_at)
+left outer join nodes on users.nodeid = nodes.id
 where
     users.service = :service
+and
+    nodes.downed != 1
 and
     replaced_at is not null and replaced_at < :timestamp
 order by
     replaced_at desc, uid desc
 limit
     :limit
 offset
     :offset
```
That is, filter out nodes.downed in the query. It also needed some help on the index hint to not choose an expensive plan.

I dunno. rfk: Thoughts on how to deal with this?

jrgm:
fixes #170 r? - @rfk 

Narrows the result set to exclude downed nodes; gives an index hint to mysql to choose an efficient plan (thanks again for the pointer).

This produces this query (mildly reformatted for readability) in stage, which can be compared to what `_GET_OLD_USER_RECORDS_FOR_SERVICE` produced:
```
SELECT
    users.uid, users.email, users.generation, users.keys_changed_at, users.client_state,
    nodes.node, nodes.downed, users.created_at, users.replaced_at
FROM users
USE INDEX (replaced_at_idx)
LEFT OUTER JOIN nodes ON users.nodeid = nodes.id
WHERE
    users.service = 1
AND
    coalesce(nodes.downed, 0) != 1
AND
    users.replaced_at IS NOT NULL
AND
    users.replaced_at < 1579578375708
ORDER BY users.replaced_at DESC, users.uid DESC
LIMIT 70072, 1000
```
The `coalesce(nodes.downed, 0) != 1` is to deal with the fact that `downed` on rows can be 0, 1 or NULL; this keeps the same semantics of `get_old_user_records`, which one existing test revealed.

Perhaps more tests could be added, but another existing test mostly covers the check that downed nodes are never removed, before or after this PR.

### DIFF
--- a/tokenserver/assignment/sqlnode/sql.py
+++ b/tokenserver/assignment/sqlnode/sql.py
@@ -13,7 +13,8 @@ import traceback
 import hashlib
 from mozsvc.exceptions import BackendError
 
-from sqlalchemy.sql import select, update, and_
+from sqlalchemy.sql import select, update, and_, desc
+from sqlalchemy.sql.functions import coalesce
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy import create_engine
 from sqlalchemy.pool import NullPool
@@ -107,25 +108,6 @@ set
 where
     email = :email
     and replaced_at is null
-""")
-
-
-_GET_OLD_USER_RECORDS_FOR_SERVICE = sqltext("""\
-select
-    uid, email, generation, keys_changed_at, client_state,
-    nodes.node, nodes.downed, created_at, replaced_at
-from
-    users left outer join nodes on users.nodeid = nodes.id
-where
-    users.service = :service
-and
-    replaced_at is not null and replaced_at < :timestamp
-order by
-    replaced_at desc, uid desc
-limit
-    :limit
-offset
-    :offset
 """)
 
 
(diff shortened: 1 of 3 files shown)


tokenserver/assignment/sqlnode/sql.py
```diff
@@ -459,16 +441,39 @@ def get_user_records(self, service, email):
...
+                        nodes.c.node,
+                        nodes.c.downed,
+                        users.c.created_at,
+                        users.c.replaced_at])
+        query = query.with_hint(users, 'USE INDEX (replaced_at_idx)',
+                                dialect_name='mysql')
+        query = query.select_from(users.outerjoin(
+            nodes,
+            users.c.nodeid == nodes.c.id))
+        query = query.where(
```
jrgm: I am **so** totally pretending to have a clue here ;-)

tokenserver/assignment/sqlnode/sql.py
```diff
@@ -459,16 +441,39 @@ def get_user_records(self, service, email):
...
+                        users.c.client_state,
+                        nodes.c.node,
+                        nodes.c.downed,
+                        users.c.created_at,
+                        users.c.replaced_at])
+        query = query.with_hint(users, 'USE INDEX (replaced_at_idx)',
+                                dialect_name='mysql')
+        query = query.select_from(users.outerjoin(
+            nodes,
+            users.c.nodeid == nodes.c.id))
```
jrgm: I am **so** totally pretending to have a clue what I'm doing here here ;-)

rfk: Heh, yep, looks like sqlalchemy magic to me 👍

jrgm:
So, I was running this in prod with a separate pool of instances, and up until this morning, things were looking pretty good. But this morning, the select queries from this purge began taking minutes to complete for reasons not yet understood. I've reduced the number of workers down to 7, and at this level, things look acceptable, but progress is slow. So, for now, I'm not going to merge this change.

jrconlin:
@jrgm: Looks like you need to squash this into a signed commit so we can merge. (github's annoying that way).

jrgm:
> @jrgm: Looks like you need to squash this into a signed commit so we can merge. (github's annoying that way).

Ah, I see. I have commit signing enabled in this repo, but the commit on the PR into this PR wasn't signed. I did `git commit --amend -c e277f8e` where e277... **was** the head of this branch (now amended), and that signed it.

This doesn't need to merge though as it doesn't make good enough progress.

jrconlin:
Huh, interesting. We were seeing a problem with github flagging another PR as not verified even though it was. @rfk do you think this might also apply to yours?

PR #173, opened 2020-02-10
fix: remember keys_changed_at when reallocating after node replacement.

rfk:
If a user's current storage node is decomissioned, they may end up
without an active `users` record in the db, and there is code to
handle this case by lazily recreating an active record on demand.

Previously, this code would only copy `generation` and `client_state`
fields, essentially forgotten any previously-seen value for the new
`keys_changed_at` field. This commit fixes things so that value is
also remembered.

I do not expect this bug to have caused any issues in practice, since
we already have a bunch of code to handle users for which we have
not yet seen a value of `keys_changed_at`. But we should remember it
when we can because it helps the server enforce that clients are
behaving correctly.

### DIFF
--- a/tokenserver/assignment/sqlnode/sql.py
+++ b/tokenserver/assignment/sqlnode/sql.py
@@ -316,7 +316,8 @@ class SQLNodeAssignment(object):
                 if cur_row.generation < MAX_GENERATION:
                     user = self.allocate_user(service, email,
                                               cur_row.generation,
-                                              cur_row.client_state)
+                                              cur_row.client_state,
+                                              cur_row.keys_changed_at)
             for old_row in old_rows:
                 # Collect any previously-seen client-state values.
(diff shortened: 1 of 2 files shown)


rfk:
Thanks @jrconlin! Weirdly, I don't seem to be able to merge this due to commit-signing, but AFAICT my commit is in fact signed. Not sure what's going on there...

jrconlin:
Yeah, I was wondering about that too, and yes, it appears verified to me as well. Looks like it's a problem for other PRs, too, so I have no idea aside from "weird github issue". Sadly, I don't have admin on this package so I can't force it.

jrconlin:
Submitted a help request with github to see what might be wrong here.

rfk:
I did a `git commit --amend` and force-pushed, let's see if that makes a difference...

rfk:
Hooray, that seemed to work \o/

PR #185, opened 2020-06-11
Release 1.5.6

rfk:


setup.py
```diff
@@ -23,7 +23,7 @@ def load_req(filename):
 
 
 setup(name='tokenserver',
-      version='1.5.3',
```
rfk, line 26: I seem to have tags for 1.5.4 and 1.5.5 so I guess maybe we missed incrementing this here?

fzzzy: I think jr may have tried to make some releases but failed because python 2 is so old now it's hard to get dependency versions that work with it. (Never thought I'd see the day!)

PR #189, opened 2020-06-20
fix: use cffi 1.13.2 because reasons

jrgm:
I noticed this message go by in a failing build:
```
  Attempting uninstall: cffi
    Found existing installation: cffi 1.13.2
    Can't uninstall 'cffi'. No files were found to uninstall.
```

... and so I tried changing that in requirements.txt to match that version, and surprise, it "works".

This PR also includes the circle-ci test from https://github.com/mozilla-services/tokenserver/pull/188 and the `docker run app:build server` shows that it starts up there.

I put this on stage, and the tokenserver daemon starts up and runs. However the event processor job fails with `/usr/bin/python: No module named zope.interface` so more work to do, but I'm going to stop for now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,6 +28,24 @@ jobs:
       - run:
           name: Test nose
           command: docker run -it app:build test_nose
+      - run:
+          name: Test that the `python` entrypoint responds
+          command: |
+            set -e
+            docker run -it app:build python --version
+      - run:
+          name: Test that the `server` entrypoint starts ok
+          command: |
+            set -e
+            container_name=$(docker run -it -e MOZSVC_SQLURI="sqlite:////tmp/tokenserver.db" --rm -d app:build server)
+            # Grab the logs now in case this crashes and is removed.
+            (docker logs --follow $container_name &)
+            sleep 10
+            # If the container above stops before running 10 seconds, the
+            # `--rm` will remove it and this `docker inspect` will exit
+            # non-zero, failing this step.
+            docker inspect -f '{{.State.Running}}' $container_name
+            docker kill $container_name
       - store_test_results:
           path: test_results
       - run:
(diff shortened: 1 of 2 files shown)


fzzzy:
The zope.interface exception sounds like something went wrong with the twisted installation. We probably need to pin to a version of twisted that supports python 2.6 syntax?

This looks good to me, r+, but I don't have permissions on this repo to actually merge it.

rfk:
> However the event processor job fails with /usr/bin/python: No module named zope.interface 

IIUC it will need to be executed using `/usr/local/bin/pypy` rather than `/usr/bin/python`; @jrconlin does that sound right to you?

Replacing `python` with `pypy` [here](https://github.com/mozilla-services/cloudops-deployment/blob/d8558a80278ca8e15fed31a7e0f06883bf213067/projects/tokenserver/puppet/modules/tokenserver/manifests/app.pp#L95)` may do the trick.

jrconlin:
Yep. Particularly with low level stuff like Zope, you need to make sure you call the correct interpreter (if I understand correctly). It *might* make sense to modify that script to use a `${python}` variable depending on if you want to reactivate the commented out block, but that's your call.

jrconlin:
@jrgm Is this resolved? Should we open a new ticket to track this problem?

jrgm:
Thanks for the pointer on running the right interpreter. I hacked up the systemd unit in stage to switch the `--entrypoint` to `pypy` and the queue task starts up now.

I'll open up another PR to make pypy exposed in docker-entrypoint.sh and do changes in cloudops-deployment to adjust the docker run command.

This current issue is resolved, although might wanna consider if any other modules should be updated.

Side note: I didn't follow the reference to `the commented out block` in ` It might make sense to modify that script to use a ${python} variable depending on if you want to reactivate the commented out block, but that's your call.` What script is that a reference to?

jrconlin:
@jrgm I was referring to [this](https://github.com/mozilla-services/cloudops-deployment/blob/d8558a80278ca8e15fed31a7e0f06883bf213067/projects/tokenserver/puppet/modules/tokenserver/manifests/app.pp#L107). It looks like this is a puppet(?) config, so I was thinking that since we've changed the interpreter we might as well make it a variable itself.

Issue #187, opened 2020-06-12
Cannot start 1.5.6 docker container (ContextualVersionConflict cffi)


jrgm:
Seeing the below exception on stage, but can also be done locally with `docker run mozilla/tokenserver:1.5.6 server`

```
Jun 12 02:52:30 ip-172-31-63-217.ec2.internal docker-tokenserver[11606]: {"v": 1, "hostname": "ip-172-31-63-217.ec2.internal", \
  "op": "gunicorn.error", "name": "gunicorn.error", "time": "2020-06-12T02:52:30.398046Z", "pid": 12, \
  "message": "Exception in worker process", \
  "error": "ContextualVersionConflict(cffi 1.13.2 (/usr/local/lib_pypy), Requirement.parse('cffi==1.12.2'), set(['tokenserver']))", \
  "traceback": "Uncaught exception:
  File \"/usr/local/site-packages/gunicorn/arbiter.py\", line 586, in spawn_worker
    worker.init_process()
  File \"/usr/local/site-packages/gunicorn/workers/base.py\", line 135, in init_process
    self.load_wsgi()
  File \"/usr/local/site-packages/gunicorn/workers/base.py\", line 144, in load_wsgi
    self.wsgi = self.app.wsgi()
  File \"/usr/local/site-packages/gunicorn/app/base.py\", line 67, in wsgi
    self.callable = self.load()
  File \"/usr/local/site-packages/gunicorn/app/wsgiapp.py\", line 50, in load
    return self.load_pasteapp()
  File \"/usr/local/site-packages/gunicorn/app/wsgiapp.py\", line 46, in load_pasteapp
    return load_pasteapp(self.cfgurl, self.relpath, global_conf=self.cfg.paste_global_conf)
  File \"/usr/local/site-packages/gunicorn/app/pasterapp.py\", line 71, in load_pasteapp
    global_conf=global_conf)
  File \"/usr/local/site-packages/paste/deploy/loadwsgi.py\", line 253, in loadapp
    return loadobj(APP, uri, name=name, **kw)
  File \"/usr/local/site-packages/paste/deploy/loadwsgi.py\", line 277, in loadobj
    global_conf=global_conf)
  File \"/usr/local/site-packages/paste/deploy/loadwsgi.py\", line 302, in loadcontext
    global_conf=global_conf)
  File \"/usr/local/site-packages/paste/deploy/loadwsgi.py\", line 326, in _loadconfig
[...]

PR #191, opened 2020-06-22
fix: allow `pypy` in docker-entrypoint.sh

jrgm:
Allow `pypy` in docker-entrypoint.sh

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -25,7 +25,7 @@ case "$1" in
             --log-config "$_SETTINGS_FILE"
         ;;
 
-    python)
+    pypy)
         exec "$@"
         ;;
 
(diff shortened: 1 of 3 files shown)


jrgm:
I added @pjenvey's suggestion about cffi from https://github.com/mozilla-services/tokenserver/issues/187#issuecomment-647706164 to this PR.

docker-entrypoint.sh
```diff
@@ -29,6 +29,10 @@ case "$1" in
         exec "$@"
```
rfk, line 29: @jrgm what do you think about either removing this existing `python` entrypoint, or changing it so that it executes `pypy` under the hood? Given the way the container is build I don't think any good can come from trying to run things in this container with `python` rather than `pypy`.

jrgm: Yes, good point. I'll remove the python entrypoint.

rfk:
@jrgm would you like a fresh tag with these fixes in?

jrgm:
Yes, a new release tag would be good.

PR #197, opened 2020-07-09
Pass a jwks key from the config through to the client. It should be a…

fzzzy:
… list of jwks key dictionaries.


rfk:
@fzzzy I've released PyFxA v0.7.6 which includes your changes. I guess we should wait until the pypy stuff is deployed before updating and merging that?

PR #200, opened 2020-07-16
Json jwks

fzzzy:
Since Paste Deploy can only pass string configuration values and not a list, make the jwks configuration parameter a serialized json object with a 'keys' key which is a list of jwks keys.


jrconlin:
Closing in favor of #201, due to unsigned commit.

PR #208, opened 2020-09-02
Pin the version of `pip` used in the `pypy` virtualenv.

rfk:
Newer versions of `pip` do not work with `pypy` v2.7 on some platforms, ref https://github.com/pypa/pip/issues/8653

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -34,7 +34,6 @@ pymysql-sa==1.0
 pyramid==1.10.4
 python-dateutil==2.8.0
 python-editor==1.0.4
-readline==6.2.4.1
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -34,7 +34,6 @@ pymysql-sa==1.0
 pyramid==1.10.4
 python-dateutil==2.8.0
 python-editor==1.0.4
-readline==6.2.4.1
```
rfk, line 37: Not entirely related to this PR, but some self-hosters have reported trouble building `readline` and I don't think it's a strictly necessary dependency.

jrconlin (approved):

PR #222, opened 2022-03-10
feat: add dryrun and limit to purge_old_records.py script

jrconlin:
## Description

add `dryrun` and `max-records` to purge_old_records script

### DIFF
--- a/tokenserver/scripts/purge_old_records.py
+++ b/tokenserver/scripts/purge_old_records.py
@@ -93,7 +93,7 @@ def purge_old_records(config_file, grace_period=-1, max_per_loop=10,
                         if not settings.dryrun:
                             backend.delete_user_record(service, row.uid)
                         counter += 1
-                    if settings.max_records:
+                    if settings and settings.max_records:
                         if counter > settings.max_records:
                             logger.info("Reached max_records, exiting")
                             return True
--- a/tokenserver/scripts/purge_old_records.py
+++ b/tokenserver/scripts/purge_old_records.py
@@ -74,8 +74,6 @@ def purge_old_records(config_file, grace_period=-1, max_per_loop=10,
                     if row.node is None:
                         logger.info("Deleting user record for uid %s on %s",
                                     row.uid, row.node)
-                        logging.info(
-                            "Deleting user records {}".format(row.uid))
                         if not settings.dryrun:
                             backend.delete_user_record(service, row.uid)
                     elif not row.downed:
@@ -89,7 +87,7 @@ def purge_old_records(config_file, grace_period=-1, max_per_loop=10,
                             backend.delete_user_record(service, row.uid)
                         counter += 1
                     elif row.downed and settings.force:
-                        logging.info(
+                        logger.info(
                             "Forcing tokenserver record delete: {}".format(
                                 row.uid
                             )
@@ -99,7 +97,7 @@ def purge_old_records(config_file, grace_period=-1, max_per_loop=10,
                         counter += 1
                     if settings.max_records:
                         if counter > settings.max_records:
-                            logging.info("Reached max_records, exiting")
+                            logger.info("Reached max_records, exiting")
                             return True
                 if len(rows) < max_per_loop:
                     break
@@ -184,7 +182,7 @@ def main(args=None):
                       help="Max records to delete")
     parser.add_option("", "--dryrun", action="store_true",
                       help="Don't do destructive things")
-    parser.add_option("", "-force", action="store_true",
+    parser.add_option("", "--force", action="store_true",
                       help="force record to be deleted from TS db,"
                       " even if node is down")
     parser.add_option("-v", "--verbose", action="count", dest="verbosity",
--- a/tokenserver/scripts/purge_old_records.py
+++ b/tokenserver/scripts/purge_old_records.py
@@ -74,17 +74,17 @@ def purge_old_records(config_file, grace_period=-1, max_per_loop=10,
                     if row.node is None:
                         logger.info("Deleting user record for uid %s on %s",
                                     row.uid, row.node)
-                        if not settings.dryrun:
+                        if not settings and settings.dryrun:
                             backend.delete_user_record(service, row.uid)
                     elif not row.downed:
                         logger.info("Purging uid %s on %s", row.uid, row.node)
                         delete_service_data(config, service, row,
                                             timeout=request_timeout,
                                             settings=settings)
-                        if not settings.dryrun:
+                        if settings and not settings.dryrun:
                             backend.delete_user_record(service, row.uid)
                         counter += 1
-                    elif row.downed and settings.force:
+                    elif row.downed and settings and settings.force:
                         logger.info(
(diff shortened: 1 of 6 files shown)


tokenserver/scripts/purge_old_records.py
```diff
@@ -66,19 +66,41 @@ def purge_old_records(config_file, grace_period=-1, max_per_loop=10,
...
+                counter = 0
                 for row in rows:
                     # Don't attempt to purge data from downed nodes.
                     # Instead wait for them to either come back up or to be
                     # completely removed from service.
                     if row.node is None:
                         logger.info("Deleting user record for uid %s on %s",
                                     row.uid, row.node)
-                        backend.delete_user_record(service, row.uid)
+                        logging.info(
```
cvalaas, line 77: isn't this logging redundant given line 75 above?

tokenserver/scripts/purge_old_records.py
```diff
@@ -155,6 +180,13 @@ def main(args=None):
                       help="Timeout for service deletion requests")
     parser.add_option("", "--oneshot", action="store_true",
                       help="Do a single purge run and then exit")
+    parser.add_option("", "--max-records", type="int", default=0,
+                      help="Max records to delete")
+    parser.add_option("", "--dryrun", action="store_true",
+                      help="Don't do destructive things")
+    parser.add_option("", "-force", action="store_true",
```
cvalaas, line 187: just one dash in force on purpose? or should it be --force?

tokenserver/scripts/purge_old_records.py
```diff
@@ -66,19 +66,39 @@ def purge_old_records(config_file, grace_period=-1, max_per_loop=10,
...
+                        if not settings.dryrun:
+                            backend.delete_user_record(service, row.uid)
                     elif not row.downed:
                         logger.info("Purging uid %s on %s", row.uid, row.node)
                         delete_service_data(config, service, row,
-                                            timeout=request_timeout)
-                        backend.delete_user_record(service, row.uid)
+                                            timeout=request_timeout,
+                                            settings=settings)
+                        logging.info(
```
cvalaas, line 84: `logger`? and is this also redundant given line 80?

tokenserver/scripts/purge_old_records.py
```diff
@@ -116,6 +136,9 @@ def delete_service_data(config, service, user, timeout=60):
     secret = tokenlib.get_derived_secret(token, secret=node_secrets[-1])
     endpoint = pattern.format(uid=user.uid, service=service, node=user.node)
     auth = HawkAuth(token, secret)
+    if settings.dryrun:
+        logging.info("Deleting", endpoint, auth, timeout)
```
cvalaas, line 140: logger?

tokenserver/scripts/purge_old_records.py
```diff
@@ -66,19 +66,39 @@ def purge_old_records(config_file, grace_period=-1, max_per_loop=10,
...
                         delete_service_data(config, service, row,
-                                            timeout=request_timeout)
-                        backend.delete_user_record(service, row.uid)
+                                            timeout=request_timeout,
+                                            settings=settings)
+                        logging.info(
+                            "Deleting user records {}".format(row.uid))
+                        if not settings.dryrun:
+                            backend.delete_user_record(service, row.uid)
+                        counter += 1
```
pjenvey, line 88: the first `if row.node is None` block doesn't inc the counter but deletes -- probably should just inc it once outside of that whole if else block instead since all the cases issue a delete.

tokenserver/scripts/purge_old_records.py
```diff
@@ -66,19 +66,39 @@ def purge_old_records(config_file, grace_period=-1, max_per_loop=10,
...
+                                row.uid
+                            )
+                        )
+                        if not settings.dryrun:
+                            backend.delete_user_record(service, row.uid)
+                        counter += 1
+                    if settings.max_records:
+                        if counter > settings.max_records:
+                            logger.info("Reached max_records, exiting")
+                            return True
```
pjenvey, line 101:
```suggestion
                            return
```

tokenserver/scripts/purge_old_records.py
pjenvey: Also missing the format specifiers
```suggestion
        logger.info("Deleting endpoint: %s auth: %s", endpoint, auth)
```
(not sure we even want the full auth output'd)

pjenvey (changes requested):

tokenserver/scripts/purge_old_records.py
jrconlin: Yeah, I was thinking that we'd only "count" rows that resulted in a deletion rather than any rows.

tokenserver/scripts/purge_old_records.py
jrconlin: the function later returns `True` or `False`, this keeps the same level of expectation.

jrconlin:
Sigh. 
No idea why the tests are failing now. 
`test_verifier_config_dynamic_issuer_discovery` is probably related to the `ValueError` work around, since requests was throwing that because the stream was already closed when the read was attempted. 

I've no idea why the `test_purge_old_records` failures are happening, since the delete is happening and returning 200, but not actually happening in the code. May look at that more later.

tokenserver/scripts/purge_old_records.py
pjenvey: The first block calls `delete_user_record` so it should inc the counter too.

pjenvey:
I was only able to reproduce that mysterious `ValueError` on a fresh virtualenv, it didn't happen on my existing virtualenv. 

It looks to be due to an issue with the `responses` lib. I think issue a `pip install -U` fixes it (not by upgrading `responses`, but because it upgrades `urllib` 🤷‍♂️ ). Alternatively, pin it to the old version and it seems to work as expected. Then you can remove the ValueError hack bits.

```diff
diff --git a/dev-requirements.txt b/dev-requirements.txt
index 191b820..6d91513 100644
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -2,5 +2,5 @@
 flake8
 mock
 nose
-responses
+responses==0.14.0
 webtest
```

jrconlin:
pinned the responses, but added comments and keeping the ValueError hack as a commented block of code. This sort of thing lives to come back from the dead and I'd like to have some record of it for the future.

pjenvey (approved):

cvalaas:
@jrconlin Do we feel like this is ready to be tested on stage and then prod (dry-run first, of course)?

jrconlin:
@cvalaas Yes. 

The new options are:
* `--max-records` Maximum number of records to delete `0` == infinite.
* `--force` Delete records from the TokenServer database, even if the corresponding Node database is unavailable (useful for very old records that point to old, invalid nodes.)
* `--dryrun` Don't do destructive things.

cvalaas:
Are you able to cut a new release? I think that'll auto-deploy to stage...

PR #230, opened 2022-06-22
fix: Detect potential loops, don't eat errors

jrconlin:
simple loop detection for purge_old_records.py

### DIFF
--- a/tokenserver/scripts/purge_old_records.py
+++ b/tokenserver/scripts/purge_old_records.py
@@ -54,6 +54,7 @@ def purge_old_records(config_file, grace_period=-1, max_per_loop=10,
     try:
         backend = config.registry.getUtility(INodeAssignment)
         patterns = config.registry['endpoints_patterns']
+        previous_list = []
         for service in patterns:
             logger.debug("Purging old user records for service: %s", service)
             # Process batches of <max_per_loop> items, until we run out.
@@ -65,6 +66,12 @@ def purge_old_records(config_file, grace_period=-1, max_per_loop=10,
                     "offset": offset,
                 }
                 rows = list(backend.get_old_user_records(service, **kwds))
+                if not rows:
+                    logger.info("No more data for %s", service)
+                    break
+                if rows == previous_list:
+                    raise Exception("Loop detected")
+                previous_list = rows.copy()
                 logger.info("Fetched %d rows at offset %d", len(rows), offset)
                 counter = 0
                 for row in rows:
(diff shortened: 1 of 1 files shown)


tokenserver/scripts/purge_old_records.py
```diff
@@ -54,6 +54,7 @@ def purge_old_records(config_file, grace_period=-1, max_per_loop=10,
     try:
         backend = config.registry.getUtility(INodeAssignment)
         patterns = config.registry['endpoints_patterns']
+        previous_list = []
```
pjenvey, line 57: This should be per service I believe (2 lines down)?

I'd argue this loop detection isn't really necessary after the last fix.

pjenvey (changes requested):

tokenserver/scripts/purge_old_records.py
jrconlin: I was debating moving it down, but then realized that we're detecting if things are the same. So any change would be fine so long as it didn't match the prior state. (And if we're doing things right, we're removing stuff along the way.)

If anything, loop detection should be nerf'd for "dry_run".

jrconlin: hrm, there's no way to provide an offset to the `backend.get_old_user_records`, so `dryrun` will always return the same list. nerf'ing isn't possible.

pjenvey: I agree it's not likely to trigger a loop detected when iterating between `services` which will have different sets of users, but it also seems confusing to me to reuse the same state across them like that.

jrconlin:
Huh, but you did get me thinking about two services that might have empty results, which would raise a false exception.

Thanks!

tokenserver/scripts/purge_old_records.py
```diff
@@ -65,6 +66,12 @@ def purge_old_records(config_file, grace_period=-1, max_per_loop=10,
                     "offset": offset,
                 }
                 rows = list(backend.get_old_user_records(service, **kwds))
+                if not rows:
+                    logger.info("No more data for %s", service)
+                    break
+                if [service].extend(rows) == previous_list:
```
pjenvey, line 72: What's `[service]` doing? `extend` going to add on inplace, always returning `None`.

`service` here is a value on the user record but it's more of a foreign key to the service table. I don't think we actually use it, it's legacy from when tokenserver was built to support multiple services.

At one point we had a 'sync-1.1' and 'sync-1.5' protocols to support but I don't think tokenserver even used this feature to support that. Long story short, consider `service` to be unrelated, separated buckets of users on potentially different services.

pjenvey (approved):

